PR #1, opened 2019-11-02
Initial implementation: supports connect and viewcal, untested

levb:
The initial server-side framework for the Microsoft Office (`msoffice`) plugin.

- A basic command interpreter, with `/msoffice connect` and `/msoffice viewcal`.
- A basic http router. Note the slightly different use patterns - http router is pre-initialized, then cloned with parameters for each request. The command router is simpler, so a new one is instantiated every request.
- A bare Microsoft Graph client.
   - Questionable choice of the `go-msgraph` package, may just implement our own client.
   - No subscriptions/webhooks yet.
- A basic OAuth2 implementation. Still need to:
   - Test/fix token expiration/refresh handling.
   - Test/fix the client secret expiration.
   - Implement `disconnect` and the various checks, consistent with other plugins.
   - Add HTML pages with bling, template support

Packaging notes:
- `./config` defines `StoredConfig` and runtime `Config`, used by most other packages. Config contains calculated fields (like `BotUserId`), and runtime dependencies (interfaces) overrideable for testing. This package also defines global constant "parameters".
- `.kvstore` - Implements the basic convenience patterns for KV store access: prefixed hashed keys, expiration, one time store - _refactor as a library_.
- `./msgraph` - The Microsoft Graph client.
- `./user` - The user object, and relevant Store interfaces.
- `./utils` - _refactor as a library_.


server/http/api_authorized.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
```
levb: This is a placeholder for a /(user-)info API, later

server/http/http.go
```diff
@@ -0,0 +1,99 @@
...
+	BotPoster         utils.BotPoster
+	IsAuthorizedAdmin func(userId string) (bool, error)
+	root              *mux.Router
+}
+
+// InitRouter initializes the router.
+func (h *Handler) InitRouter() {
+	h.root = mux.NewRouter()
+	api := h.root.PathPrefix("/api/v1").Subrouter()
+	api.Use(authorizationRequired)
```
mickmister: :+1: for this pattern

server/kvstore/ots.go
```diff
@@ -0,0 +1,37 @@
...
+
+package kvstore
+
+import (
+	"time"
+
+	"github.com/mattermost/mattermost-server/plugin"
+)
+
+// OneTimeStore is a KV store that deletes each record after the first load,
```
mickmister: :+1:

server/plugin/authorization.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return false, err
+	}
+	if strings.Contains(user.Roles, "system_admin") { // || p.userAuthorized(mattermostID) {
+		return true, nil
+	}
+	return false, nil
+}
+
+// func (p *Plugin) userAuthorized(userName string) bool {
```
mickmister: Is the commented out code in this file intentionally left here for future use, or leftover debug code?

server/plugin/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+
+func NewWithConfig(conf *config.Config) *Plugin {
+	return &Plugin{
+		configLock: &sync.RWMutex{},
+		config:     conf,
+	}
+}
+
+func (p *Plugin) OnActivate() error {
+	// if p.Templates == nil {
```
mickmister: Same question about commented code

server/plugin/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	botUserId := conf.BotUserId
+	if newStored.BotUserName != oldStored.BotUserName {
+		user, appErr := p.API.GetUserByUsername(newStored.BotUserName)
+		if appErr != nil {
+			return errors.WithMessage(appErr, fmt.Sprintf("unable to load user %s", newStored.BotUserName))
+		}
+		botUserId = user.Id
+	}
+
+	mattermostSiteURL := *p.API.GetConfig().ServiceSettings.SiteURL
```
mickmister: I'm concerned about this error'ing out if the `Site URL` is not set, with no appropriate error shown.

server/user/oauth2_store.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func (s *oauth2StateStore) Verify(state string) error {
+	data, err := s.kv.Load(state)
+	if err != nil {
+		return err
+	}
+	if string(data) != state {
+		return errors.New("authentication attempt expired, please try again.")
```
mickmister: What is being verified here?

server/user/user_store.go
```diff
@@ -0,0 +1,71 @@
...
+		kv: kvstore.NewHashedKeyStore(s, UserKeyPrefix),
+	}
+}
+
+func (s *store) LoadRemoteUser(mattermostUserId string, ref interface{}) error {
+	return kvstore.LoadJSON(s.kv, mattermostUserId, ref)
+}
+
+func (s *store) LoadMattermostUserId(remoteUserId string) (string, error) {
+	data, err := s.kv.Load(remoteUserId)
```
mickmister: Do we want to prefix the `remoteUserId` with something, for the keys in the kv store? Along with every other user-specific key

server/utils/bot_poster.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/mattermost/mattermost-server/mlog"
+	"github.com/mattermost/mattermost-server/model"
+	"github.com/mattermost/mattermost-server/plugin"
+
+	"github.com/mattermost/mattermost-plugin-msoffice/server/config"
+)
+
+type BotPoster interface {
+	PostDirect(userID, message, postType string) error
+	PostEphemeral(userID, channelId, message string)
```
mickmister: Is there a use case where we post a publicly in a channel? We can just add it when we need it.

server/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"net/url"
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func NormalizeInstallURL(mattermostSiteURL, jiraURL string) (string, error) {
```
mickmister: 0/5 for what to change it to

```suggestion
func NormalizeInstallURL(mattermostSiteURL, microsoftURL string) (string, error) {
```

server/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+	if u.Host == "" {
+		return "", errors.Errorf("Invalid URL, no hostname: %q", jiraURL)
+	}
+	if u.Scheme == "" {
+		u.Scheme = "https"
+	}
+
+	jiraURL = strings.TrimSuffix(u.String(), "/")
+	if jiraURL == strings.TrimSuffix(mattermostSiteURL, "/") {
+		return "", errors.Errorf("%s is the Mattermost site URL. Please use your Jira URL with `/jira install`.", jiraURL)
```
mickmister: Change to include Microsoft terms instead of Jira terms

server/user/user_store.go
levb: It is, see https://github.com/mattermost/mattermost-plugin-msoffice/pull/1/files#diff-4cd1f48867ecea93ce7781326fe70ca5R25

server/user/oauth2_store.go
levb: Looking at refactoring OAuth to properly use refresh tokens, so will review this comment as part of that.

levb:
@crspeller RE: ...utils stuff could be better placed. Some of it should be upgraded to MM server helpers - do you mind filing github issues on these? I'll work them into HW.

PR #2, opened 2019-11-12
Add mocks and unit tests to http and msgraph packages

### LINKED ISSUE #3: Use github.com/stretchr/testify/mock for unit tests
The recommended best practice at Mattermost is to use `github.com/stretchr/testify/mock` to mock interfaces in unit tests. In other plugins, like Jira, we had been constrained by the flat plugin package layout, and the existing data structures made it difficult to break it into subpackages.

This plugin was set up to avoid these issues, and the functionality is more carefully packaged, so using `github.com/stretchr/testify/mock` should be easy(-er).

cpurta:
Added a mocks package that mocks out the functionality of some the
interfaces needed by the oauth handlers. This allows testing of each
error check within the oauth handler functions. I also added
github.com/jarcoal/httpmock as a dependecy to test both the oauth code
to token exchange and each of the calls that will be made out to the
microsoft graph API.

The tests cover how all errors are encountered for the oauth connection.
They also ensure that access token refreshing happens when the token is
expired. This tests to make sure the golang oauth client works out of
the box for this plugin.

This will also resolve #3


server/http/http.go
```diff
@@ -21,6 +21,7 @@ type Handler struct {
 	Config            *config.Config
 	UserStore         user.Store
 	API               plugin.API
+	OAuth2StateStore  user.OAuth2StateStore
```
levb: 👍 - exactly.

server/http/oauth2_complete_test.go
```diff
@@ -0,0 +1,270 @@
+package http
```
levb: 0/5 I have not looked at every detail yet, but I wonder if the tests can be made 100% external to the package and moved into a sub-package (`test_http`?), but only if it doesn't create hurdles in testing.

server/http/oauth2_complete_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/jarcoal/httpmock"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/config"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/mocks"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/user"
+	"github.com/mattermost/mattermost-server/app"
+	"github.com/stretchr/testify/assert"
+)
+
+func getHTTPRequest(userID, rawQuery string) *http.Request {
```
levb: 0/5 suggest `s/getHTTPRequest/userRequest/`

server/http/oauth2_complete_test.go
```diff
@@ -0,0 +1,270 @@
...
+		RawQuery: rawQuery,
+	}
+	r.Header.Add("Mattermost-User-ID", userID)
+
+	return r
+}
+
+func TestOauth2Complete(t *testing.T) {
+	api := &app.PluginAPI{}
+	kv := &mocks.MockKVStore{
```
levb: Would you be open, either as a part of this PR, or as a separate one, to instrument all interface mocks to use `github.com/stretchr/testify/mock` like we do elsewhere? I will file a separate issue for it, and explain the history/context there.

server/http/oauth2_complete_test.go
```diff
@@ -0,0 +1,270 @@
...
+		KVStore: make(map[string][]byte),
+		Err:     nil,
+	}
+
+	config := &config.Config{}
```
levb: 1/5 use a go-idiomatic struct literal?

server/http/oauth2_complete_test.go
levb: https://github.com/mattermost/mattermost-plugin-msoffice/issues/3

levb:
Also @cpurta please sign the CLA in the link above.

levb:
@cpurta `dev` has been merged, so you can re-base/re-point to `master` now.

server/http/oauth2_complete_test.go
cpurta: I attempted to do that when I was first writing the code. Since the OAuth fields are stored in the StoredConfig which is embedded in the Config struct I am unable to directly assign those in the struct literal. When I do I get the following error:
```
cannot use promoted field StoredConfig.OAuth2Authority in struct literal of type Config
```
We can get around that if the StoredConfig is not embedded in the Config struct.

server/http/oauth2_complete_test.go
cpurta: I can do that. I think that it would probably be best to wrap that into this PR.

cpurta:
Signed the CLA yesterday and now seeing the bot recognize that. Also just merge the `master` into my branch and everything is looking up to date.

server/http/oauth2_complete_test.go
cpurta: I can work on doing that. I assume if that works without too many hurdles we should the same thing for the tests in the `msgraph` package?

server/http/oauth2_complete_test.go
cpurta: Renamed that function and all calls to it

server/http/oauth2_complete_test.go
levb: Fair enough, 👍

server/http/test_http/oauth2_complete_test.go
```diff
@@ -0,0 +1,294 @@
...
+    "userPrincipalName": "userPrincipalName-value",
+    "id": "id-value"
+}`)
+
+	httpmock.RegisterResponder("GET", meRequestURL, meResponder)
+}
+
+func getMockBotPoster(ctrl *gomock.Controller, err error) utils.BotPoster {
+	m := mock_utils.NewMockBotPoster(ctrl)
+	m.EXPECT().PostDirect(gomock.Any(), gomock.Any(), gomock.Any()).Return(err).AnyTimes()
```
cpoile: I think putting the mock Expects into a function like this seems good at first, but then I realized that you're not actually testing whether the post was called or not (because of `AnyTimes`). I think that goes for all the mocks. You might have to redesign the tests a bit so that you can specify exactly what each mock is called with (to verify that they are being called with the expected arguments -- right now you're using `gomock.Any()`) and how many times they were called (so that the tests error out when the mock is /not/ called, or called more than you expected). Right now, the botPoster could be called zero times, and the test would still pass, right?

cpurta: That is correct. When I was writing those functions it did feel like a bit of a cheat to just accept anything and just let it get calls any number of times. I have figured out a solution to build the mocks out with expected parameters and max number of times for each call. I will be pushing that soon.

server/http/test_http/oauth2_complete_test.go
```diff
@@ -271,24 +313,83 @@ func statusOKGraphAPIResponderFunc() {
...
 }
 
-func getMockBotPoster(ctrl *gomock.Controller, err error) utils.BotPoster {
+func getBotPosterMessage(displayName string) string {
+	return fmt.Sprintf("### Welcome to the Microsoft Office plugin!\n"+
+		"Here is some info to prove we got you logged in\n"+
+		"Name: %s \n", displayName)
+}
+
+type mockBotPosterConfig struct {
```
cpoile: Hmm, it's an interesting approach. But you probably realized that you're spending a lot of time handling the logic of the mocks, and then you'd have to add more logic if you want to do something else with the mocks. 

I probably would have gone with a `setupMocks(mock1, mock2, ...etc)` function that you could define for each test case, which would do the setups for the mocks for that particular case. That way you don't have to worry about config structs at all. The main test would call each test case's `setupMocks` function to prepare the mocks before making the call. Make sense?

server/http/test_http/oauth2_complete_test.go
```diff
@@ -271,24 +313,83 @@ func statusOKGraphAPIResponderFunc() {
...
+func getMockKVStore(ctrl *gomock.Controller, config *mockKVStoreConfig) kvstore.KVStore {
 	m := mock_kvstore.NewMockKVStore(ctrl)
-	m.EXPECT().Load(gomock.Any()).Return(bytes, err).AnyTimes()
-	m.EXPECT().Store(gomock.Any(), gomock.Any()).Return(err).AnyTimes()
-	m.EXPECT().Delete(gomock.Any()).Return(err).AnyTimes()
+	m.EXPECT().Load(config.loadKey).Return(config.loadBytes, config.loadErr).MaxTimes(config.maxTimesLoad)
+
+	if config.useStoreAnyValue {
+		if config.useStoreAnyKey {
+			m.EXPECT().Store(gomock.Any(), gomock.Any()).Return(config.storeErr).MaxTimes(config.maxTimesStore)
```
cpoile: Are you sure you want `MaxTimes`, and not just `Times`?

cpoile:
@cpurta Also, as soon as you want to let us know the PR is ready for another review, hit the `refresh` symbol next to our names in the top right -- that will ping us through the Mattermost Github plugin that it's time to take another look. :)

server/http/test_http/oauth2_complete_test.go
cpurta: Ahh... Yeah we probably want to just use `Times` in this case. I must have missed that in their docs and just went straight to using `MaxTimes`.

server/http/test_http/oauth2_complete_test.go
cpurta: That makes sense. I didn't think about passing in the mocks to a function for setup. That would make it much more clean than a config and doing logical steps for different configurations.

server/http/test_http/oauth2_complete_test.go
```diff
@@ -0,0 +1,395 @@
...
+    }
+}`)
+
+	httpmock.RegisterResponder("GET", meRequestURL, meResponder)
+}
+
+func statusOKGraphAPIResponderFunc() {
+	tokenURL := "https://login.microsoftonline.com/common/oauth2/v2.0/token"
+
+	tokenResponder := httpmock.NewStringResponder(http.StatusOK, `{
```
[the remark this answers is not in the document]
cpurta: Can do!

server/http/testhttp/oauth2_complete_test.go
```diff
@@ -55,51 +53,59 @@ func TestOAuth2Complete(t *testing.T) {
 		handler               shttp.Handler
 		r                     *http.Request
 		w                     *mockResponseWriter
+		mockKVStore           *mock_kvstore.MockKVStore
+		mockOAuth2StateStore  *mock_user.MockOAuth2StateStore
+		mockBotPoster         *mock_utils.MockBotPoster
```
cpoile: Since you pass these into the `setupMocks` fn, you don't need to store each of these in each test case -- just create them in the body of the test (right before 241). 

Basically you only need to store things in the test case that are going to be different between test cases. So can you think of anything else that you're storing (and repeating for every test case) that you don't need to?

server/http/testhttp/oauth2_connect_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	tcs := []struct {
+		name                 string
+		handler              shttp.Handler
+		r                    *http.Request
+		w                    *mockResponseWriter
+		mockKVStore          *mock_kvstore.MockKVStore
+		mockOAuth2StateStore *mock_user.MockOAuth2StateStore
+		mockBotPoster        *mock_utils.MockBotPoster
+		setupMocks           func(*mock_kvstore.MockKVStore, *mock_user.MockOAuth2StateStore, *mock_utils.MockBotPoster)
```
cpoile: (same as prev comment)

cpoile: (I didn't mean to highlight line 30 here, oops)

server/http/testhttp/oauth2_complete_test.go
cpurta: Ok. That makes sense. Definitely no point in storing that in the struct since those never change.

server/http/testhttp/oauth2_complete_test.go
```diff
@@ -0,0 +1,316 @@
...
+	}{
+		{
+			name: "unauthorized user",
+			handler: shttp.Handler{
+				Config: config,
+				API:    api,
+			},
+			setupMocks: func(kv *mock_kvstore.MockKVStore, ss *mock_user.MockOAuth2StateStore, bp *mock_utils.MockBotPoster) {},
+			r:          &http.Request{},
+			w:          defaultMockResponseWriter(),
```
cpoile: Same for this

server/http/testhttp/oauth2_complete_test.go
```diff
@@ -0,0 +1,316 @@
...
+		r                     *http.Request
+		w                     *mockResponseWriter
+		setupMocks            func(*mock_kvstore.MockKVStore, *mock_user.MockOAuth2StateStore, *mock_utils.MockBotPoster)
+		registerResponderFunc func()
+		expectedHTTPResponse  string
+		expectedHTTPCode      int
+	}{
+		{
+			name: "unauthorized user",
+			handler: shttp.Handler{
```
cpoile: Can move this one to the test body as well.

server/http/testhttp/oauth2_connect_test.go
```diff
@@ -0,0 +1,87 @@
...
+		r                    *http.Request
+		w                    *mockResponseWriter
+		setupMocks           func(*mock_kvstore.MockKVStore, *mock_user.MockOAuth2StateStore, *mock_utils.MockBotPoster)
+		expectedHTTPResponse string
+		expectedHTTPCode     int
+	}{
+		{
+			name: "unauthorized user",
+			r:    &http.Request{},
+			w:    defaultMockResponseWriter(),
```
cpoile: Same

server/http/testhttp/oauth2_connect_test.go
```diff
@@ -0,0 +1,87 @@
...
+			handler: shttp.Handler{
```
cpoile: Same

cpoile:
@levb Final pass?

PR #8, opened 2019-12-05
MM-19895: Added event notification PostActions

levb:


server/api/notification.go
```diff
@@ -184,49 +194,153 @@ func (h *notificationHandler) processNotification(n *remote.Notification) error
...
+			URL:     h.actionURL(config.PathRespond),
+			Context: context,
+		},
+	}
+
+	for _, o := range []string{OptionNotResponded, OptionYes, OptionNo, OptionMaybe} {
+		pa.Options = append(pa.Options, &model.PostActionOptions{Text: o, Value: o})
+	}
+	switch event.ResponseStatus.Response {
+	case "notResponded":
```
mickmister: Not blocking but can these be constantized?

levb: Will do separately, more changes are likely to come to that interface.

PR #9, opened 2019-12-07
Added tee log to admins as bot messages

levb:
This addition eases the debugging, especially in HA/kubernetes environments, and should be useful for customer support, as well as development. Note that when turned to `debug`, it will post debug log entries even if the overall server log level is lower.

cc @hanzei to consider as a `Helpers` feature.

- Also pulled in `Logger` and `IsUserAdmin` into `utils/bot`.
- Also updated `.cicrleci` scripts, the build now works. Fixes https://mattermost.atlassian.net/browse/MM-19883.

Details:
Adds config settings to turn teeing on/off and control its level and verbosity.
image

With these settings, the output of `/msoffice subscribe` looks like this,
image


server/utils/bot/poster.go
```diff
@@ -5,63 +5,71 @@ package bot
...
+	post.UserId = bot.mattermostUserID
+	if _, err := bot.pluginAPI.CreatePost(post); err != nil {
 		return err
 	}
 	return nil
 }
 
-func (poster *poster) PostEphemeral(userId, channelId, message string) {
+// DM posts a simple Direct Message to the specified user
+func (bot *bot) dmAdmins(format string, args ...interface{}) error {
```
mickmister: Do we want this send the debug messages to all admins?

server/plugin/authorization.go
```diff
@@ -5,26 +5,20 @@ package plugin
 
 import (
 	"strings"
+
+	"github.com/mattermost/mattermost-plugin-msoffice/server/utils/bot"
 )
 
 // isAuthorized returns true if the user is authorized to use the workflow plugin's admin-level APIs/commands.
```
cpoile: workflow plugin? :)

server/plugin/authorization.go
```diff
@@ -5,26 +5,20 @@ package plugin
...
-	list := strings.Split(p.getConfig().AdminUserIDs, ",")
-	for _, u := range list {
-		if mattermostID == strings.TrimSpace(u) {
-			return true
-		}
-	}
-	return false
+	conf := p.getConfig()
+	bot := bot.NewBot(p.API, conf.BotUserID).WithConfig(conf.BotConfig)
+	return bot.IsUserAdmin(mattermostUserID), nil
```
cpoile: Sorry, I'm confused about this... Why would IsUserAdmin belong in the bot namespace? We're creating a newbot just to call a utility function, which deals with authorization -- shouldn't it just be here like before?

server/plugin/http/oauth2_complete_test.go
```diff
@@ -121,9 +120,11 @@ func TestOAuth2Complete(t *testing.T) {
...
 				us.EXPECT().StoreUser(gomock.Any()).Return(nil).Times(1)
 				ss.EXPECT().VerifyOAuth2State(gomock.Eq("user_fake@mattermost.com")).Return(nil).Times(1)
-				poster.EXPECT().PostDirectf(
-					gomock.Eq("fake@mattermost.com"), gomock.Eq(getBotPosterMessage("displayName-value"))).
-					Return(nil).Times(1)
+				poster.EXPECT().DM(
+					gomock.Eq("fake@mattermost.com"),
+					gomock.Eq(api.WelcomeMessage),
+					gomock.Eq("displayName-value"),
+				).Return(nil).Times(1)
```
cpoile: 👍

server/utils/bot/logger.go
```diff
@@ -0,0 +1,98 @@
...
+
+func toKeyValuePairs(in map[string]interface{}) (out []interface{}) {
+	for k, v := range in {
+		out = append(out, k)
+		out = append(out, v)
+	}
+	return out
+}
+
+func (bot *bot) With(logContext LogContext) Logger {
```
cpoile: It simplifies writing out kv pairs to the logging messages, right? This is cool, I like it.

server/plugin/authorization.go
levb: sorry, what's the question? I most definitely copied many things from the workflow plugin.

server/plugin/authorization.go
levb: Well, the idea is that these users are "bot admins". As such, they may receive plugin's log messages, with this PR. They also get a (simple) check for being able to administer other plugin functions.

server/utils/bot/logger.go
levb: Yes, it also allows to (a) easily combine it with the printf syntax for the actual messages, and (b) cascade contexts by adding more context to a (local) logger in each function down the stack.

server/plugin/authorization.go
levb: Ah, nm!

server/utils/bot/poster.go
levb: For now, the messages just go to all admins; it may be better to make a configurable channel to send them to (as an alternative). Also, we should make the admin list accept usernames. The log level for these "tee-ed " messages can be controlled in the config.

server/plugin/authorization.go
levb: Also, maybe the names `bot` and `NewBot` are partly to blame for the confusion here, not intention-revealing enough. `BotAPI`?

cpoile: `BotAPI`, or `GetBot` maybe? `NewBot` definitely through me for a second.

levb: I like `GetBot` - doesn't require another "API", yet reveals the light-weight nature of the call, hopefully. I also really like how `bot.DM(...)` looks, would like to keep `bot`.

levb: @cpoile now better?

cpoile: 👍

PR #10, opened 2019-12-09
Only prepend baseURL to relative paths

jfrerich:
#Summary

This PR fixes an issue where full paths, that already contained the baseURL, were getting prepended with the baseURL.  Now, relative paths are determined by looking at the path string and checking that the first character is a `/`.

Also fixed an issue where inBody was not getting set to the buf value, thus alway `nil`.

### DIFF
--- a/server/remote/msgraph/call.go
+++ b/server/remote/msgraph/call.go
@@ -22,10 +22,11 @@ func (c *client) Call(method, path string, in, out interface{}) (responseData []
 	if err != nil {
 		return nil, errors.WithMessage(err, errContext)
 	}
-	if len(path) > 0 && path[0] != '/' {
-		path = "/" + path
+
+	// prepend baseURL to relative paths
+	if len(path) > 0 && path[0] == '/' {
+		path = baseURL.String() + path
 	}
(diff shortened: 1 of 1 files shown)


server/remote/msgraph/call.go
```diff
@@ -34,6 +35,7 @@ func (c *client) Call(method, path string, in, out interface{}) (responseData []
 		if err != nil {
 			return nil, err
 		}
+		inBody = buf
```
mickmister: What was this `inBody` variable used for before if it wasn't getting assigned?

jfrerich: @mickmister, it wasn't being used.  The `Call()` method was written but not yet thoroughly tested and the use cases didn't actually leverage this code.  An example would be a call to this function that didn't require a JSON payload.  

I ran into an issue when using is this function for creating a calendar event which is a POST method. Other calls to the function, thus far, are GET methods and didn't require a JSON Payload.

levb: Was my bad - never tested this path, it's all POC. Jason discovered it trying to use .Call

PR #16, opened 2019-12-13
Fixed relative path handling in msgraph.Call

levb:


server/remote/msgraph/call.go
```diff
@@ -18,13 +18,20 @@ import (
...
-	// prepend baseURL to relative paths
-	if len(path) > 0 && path[0] == '/' {
+	if pathURL.Scheme == "" || pathURL.Host == "" {
+		var baseURL *url.URL
+		baseURL, err = url.Parse(c.rbuilder.URL())
+		if err != nil {
+			return nil, errors.WithMessage(err, errContext)
+		}
+		if path[0] != '/' {
+			path = "/" + path
```
mickmister: Not a big deal, but do we want to reassign a function parameter value like this?

PR #18, opened 2019-12-19
[MM-19893] Outlook user status sync intial implementation

mickmister:
#### Summary

This PR implements the user status sync feature, between the Outlook Calendar and Mattermost.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-19893


server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+
+	_, err := j.api.GetUserAvailability()
+	if err != nil {
+		log.Errorf("Error during Availability job", "error", err.Error())
+	}
+
+	log.Debugf("Availability job finished")
+}
+
+func (api *api) GetUserAvailability() (string, error) {
```
mickmister: This is called by a single user using `/availability`, possibly just for debugging?

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	if len(sched) == 0 {
+		return "No schedule info found", nil
+	}
+
+	s := sched[0]
+	av := s.AvailabilityView[0]
+	return api.setUserStatusFromAvailability(api.mattermostUserID, av), nil
+}
+
+func (api *api) GetAllUsersAvailability() (string, error) {
```
mickmister: This is called by the recurring job. For debugging in development, you can trigger it with `/availability all`

server/job/recurring_job.go
```diff
@@ -0,0 +1,5 @@
+package job
+
+type RecurringJob interface {
+	Run()
```
mickmister: This file seems pretty useless as it stands

server/plugin/plugin.go
```diff
@@ -67,10 +67,15 @@ func (p *Plugin) OnActivate() error {
...
 
 	p.httpHandler = http.NewHandler()
-	p.notificationHandler = api.NewNotificationHandler(p.newAPIConfig())
+
+	conf := p.newAPIConfig()
+	p.notificationHandler = api.NewNotificationHandler(conf)
 
 	command.Register(p.API.RegisterCommand)
 
+	// j := api.NewAvailabilityJob(api.New(conf, ""))
```
mickmister: Job is currently disabled. It should be easier for us to configure on and off.

server/remote/common.go
```diff
@@ -19,7 +19,9 @@ const RFC3339NanoNoTimezone = "2006-01-02T15:04:05.999999999"
 func NewDateTime(t time.Time) *DateTime {
 	return &DateTime{
 		DateTime: t.Format(RFC3339NanoNoTimezone),
-		TimeZone: t.Format("MST"),
+		// TimeZone: t.Format("MST"),
+		TimeZone: "Eastern Standard Time",
```
mickmister: Microsoft requires the full name spelled out. We should use the current timezone of the server, I believe.

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+	EndTime   *remote.DateTime `json:"endTime"`
+
+	// Size of each chunk of time we want to check
+	// This can be equal to end - start if we want, or we can get more granular results by making it shorter.
+	// For the graph API: The default is 30 minutes, minimum is 6, maximum is 1440
+	// 15 is currently being used on our end
+	AvailabilityViewInterval int `json:"availabilityViewInterval"`
+}
+
+func (c *client) GetSchedule(remoteUserID string, schedules []string, startTime, endTime *remote.DateTime, availabilityViewInterval int) ([]*remote.ScheduleInformation, error) {
```
mickmister: Even though we are using app-level creds, we still need to provide a user id to the request. The body of the request contains several users' information, but the url needs to have one in it as well.

Our options are
```
POST /me/calendar/getSchedule 
POST /users/{id|userPrincipalName}/calendar/getSchedule
```

`me` doesn't work for the app-level user, and I don't think it has a user id. This endpoint specifically supports app-level users, so I'm not sure why we need to provide a user id in the url, or what the point of the of id is if the users we are searching for are in the body.

https://docs.microsoft.com/en-us/graph/api/calendar-getschedule

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, r := range sorted {
+		for _, sched := range r.Body.Value {
+			result = append(result, sched)
+		}
+	}
+
+	return result, nil
+}
+
+func getFullBatchRequest(remoteUserID string, schedules []string, params *GetScheduleRequest) []*SingleRequest {
```
mickmister: This function needs to dynamically create the batch requests based on the length of users provided.

server/remote/msgraph/remote.go
```diff
@@ -47,6 +48,29 @@ func (r *impl) NewClient(ctx context.Context, token *oauth2.Token) remote.Client
 	return c
 }
 
+// NewAppLevelClient creates a new client used for app-only permissions.
+func (r *impl) NewAppLevelClient(ctx context.Context) remote.Client {
+	httpClient := &http.Client{}
+
+	// Hack. getAppLevelToken should just use http and not a client
+	c := &client{
```
mickmister: Need to manage app-level user's oauth token fetching better

plugin.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "com.mattermost.msoffice",
-    "name": "TODO:name",
-    "description": "TODO: description.",
+    "name": "Microsoft Calendar",
```
levb: 👍 I again suggest that we rename the repo before too late to `...-mscalendar`

server/api/api.go
```diff
@@ -65,6 +69,7 @@ type Dependencies struct {
 	Poster            bot.Poster
 	Remote            remote.Remote
 	IsAuthorizedAdmin func(userId string) (bool, error)
+	API               plugin.API
```
levb: I am not keen on passing and exposing the entire p.API everywhere, 1/5 would like to keep it isolated. I did not solve that in the POC since there were few dependencies other than the `Poster`/`Logger`, but in my other project,started off the same code base here is how I solved it. https://github.com/levb/mattermost-plugin-solar-lottery/blob/master/server/api/api.go#L22

server/api/api.go
```diff
@@ -96,6 +101,10 @@ func (api *api) MakeClient() (remote.Client, error) {
 	return api.Remote.NewClient(context.Background(), api.user.OAuth2Token), nil
 }
 
+func (api *api) MakeAppClient() (remote.Client, error) {
```
levb: 0/5 somehow I find `App` to be insufficiently intention-revealing here (same in Jira, ``MakeClientForServer` or something? - no good). `MakeSuperuserClient` or at least `MakeAppLevelClient` to be consistent with `remote`.

Also 0/5 suggest `s/Make/New/`, and removing the unused error return?

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/robfig/cron/v3"
+
+	"github.com/mattermost/mattermost-plugin-msoffice/server/job"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/remote"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/utils"
+	"github.com/mattermost/mattermost-plugin-msoffice/server/utils/bot"
+)
+
+const (
+	AVAILABILITY_VIEW_FREE              = '0'
```
levb: 2/5 let's use the Go style (CamelCase) in this repo

server/api/api.go
mickmister: I agree, and meant to put this as one of the todos

server/remote/common.go
levb: Timezone handling can be a thorny issue for a calendaring app. 1/5 I suggest doing everything in hardcoded UTC for the time being, and file a separate Jira ticket to POC user timezone handling. Some things we should keep in mind:
1. Seems like the overall approach should be storing (encoded) time.Time values and converting them to local _user_-specific display strings as needed.
2. Submitting the time values to Microsoft, we should be careful to accurately encode/decode time's `Location` to preserve the original timezone. This is important for recurring events, where timezones are affected by daylight time changes differently.
3. It'll likely be required at some point to display "other users" local times (i.e. meeting time as it appears to `@other-user`.

server/api/availability.go
levb: Suggest a different name, since this actually _Sets_ the user status.

2/5 `api.SyncStatusForSingleUser` and `api.SyncStatusForUsers`. I am not really concerned about long names for high-level `api` methods since they tend to be called rarely, once/user action. Being intention-revealing is important for those.

0/5 I'd consider restructuring the code around (a) getting the desired status, and (b) setting it, for 1 and a collection of users (whether it's "all" or a subset should be up to the caller).

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	AVAILABILITY_VIEW_TENTATIVE         = '1'
+	AVAILABILITY_VIEW_BUSY              = '2'
+	AVAILABILITY_VIEW_OUT_OF_OFFICE     = '3'
+	AVAILABILITY_VIEW_WORKING_ELSEWHERE = '4'
+)
+
+type availabilityJob struct {
+	api API
```
levb: I tried to avoid this pattern in `notification.go` - the thinking was that `api` is for an "acting user", has (at least) the current user's MattermostUserID etc. associated with it.

As I was working on solar lottery, i realized that this totally makes sense, because the `api` should just be operating in the context of the bot user. In fact, I am thinking that maybe the Bot User is the right concept to use for the "application-level" client. Can talk offline about it, no immediate change request here.

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func NewAvailabilityJob(api API) job.RecurringJob {
+	return &availabilityJob{api: api}
+}
+
+func (j *availabilityJob) Run() {
+	c := cron.New()
```
levb: Need (a separate ticket) for HA-proofing this by using a distributed lock to run the cron job on one server only. (failover??? does the lock expire?)

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	c.AddFunc("* * * * *", j.Work)
+	c.Start()
+}
+
+func (j *availabilityJob) getLogger() bot.Logger {
+	return j.api.(*api).Logger
```
levb: Doesn't API implement `bot.Logger`? If not, 2/5 it should, see https://github.com/levb/mattermost-plugin-solar-lottery/blob/a37a0ba0512bad58ced4a542740ec7cf9b53aa3c/server/api/api.go#L20 for how I did it there, some improvements to be back-ported.

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+		return "No schedule info found", nil
+	}
+
+	s := sched[0]
+	av := s.AvailabilityView[0]
+	return api.setUserStatusFromAvailability(api.mattermostUserID, av), nil
+}
+
+func (api *api) GetAllUsersAvailability() (string, error) {
+	client, err := api.MakeAppClient()
```
levb: See my comment about using the Bot Account, and maybe requiring that it is adequately mapped to a specific Graph user (as all normal user accounts are?)

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+		return "", err
+	}
+
+	if len(users) == 0 {
+		return "No connected users found", nil
+	}
+
+	scheduleIDs := []string{}
+	for _, u := range users {
+		scheduleIDs = append(scheduleIDs, u.Email)
```
levb: just confirming, u.Email is the Graph-side user's email, not what Mattermost knows to be the user's email.

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	start, end, timeWindow := getTimeInfoForAvailability()
+
+	sched, err := client.GetSchedule(users[0].RemoteID, scheduleIDs, start, end, timeWindow)
+	if err != nil {
+		return "", err
+	}
+
```
levb: nit: remove blank line, it's 1 "paragraph"?

server/api/availability.go
```diff
@@ -0,0 +1,167 @@
...
+	if res != "" {
+		return res, nil
+	}
+
+	return utils.JSONBlock(sched), nil
+}
+
+func getTimeInfoForAvailability() (start, end *remote.DateTime, timeWindow int) {
+	start = remote.NewDateTime(time.Now())
+	end = remote.NewDateTime(time.Now().Add(15 * time.Minute))
```
levb: 1/5 `const` it (15 minutes)

server/plugin/plugin.go
levb: 2/5 The job needs to be notified (or restarted) on configuration changes

server/remote/msgraph/call.go
```diff
@@ -36,29 +37,44 @@ func (c *client) Call(method, path string, in, out interface{}) (responseData []
 	}
 
 	var inBody io.Reader
+	var contentType string
```
levb: 0/5 I like what you did, but I wonder if it'd simplify (and improve the functionality) of the code to have contentType passed in by the caller. If `.Call` gets used widely, we can always add a `.CallJSON` to simplify the caller-side code.

server/remote/msgraph/call.go
```diff
@@ -36,29 +37,44 @@ func (c *client) Call(method, path string, in, out interface{}) (responseData []
...
+		req.Header.Add("Content-Type", contentType)
 	}
+
 	if c.ctx != nil {
 		req = req.WithContext(c.ctx)
 	}
 
-	resp, err := c.httpClient.Do(req)
+	httpClient := c.httpClient
+	if httpClient == nil {
```
levb: why?

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+
+type GetScheduleRequest struct {
+	// List of emails of users that we want to check
+	Schedules []string `json:"schedules"`
+
+	// Overall start and end of entire search window
+	StartTime *remote.DateTime `json:"startTime"`
+	EndTime   *remote.DateTime `json:"endTime"`
+
+	// Size of each chunk of time we want to check
```
levb: nit: format the comment

server/remote/msgraph/get_schedule_batched.go
levb: See my comment about using (and mapping) the BotAccount. We can still use special-case the credentials to invoke graph, but require that a user is specified.

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+		p := &GetScheduleRequest{
+			Schedules:                schedules, // need to chunk these out
+			StartTime:                params.StartTime,
+			EndTime:                  params.EndTime,
+			AvailabilityViewInterval: params.AvailabilityViewInterval,
+		}
+		req := &SingleRequest{
+			URL:    u,
+			Method: "POST",
+			Body:   p,
```
levb: 1/5 would read better if Body initialization was embedded.

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return result, nil
+}
+
+func getFullBatchRequest(remoteUserID string, schedules []string, params *GetScheduleRequest) []*SingleRequest {
+	u := "/Users/" + remoteUserID + "/calendar/getSchedule"
+
+	makeRequest := func() *SingleRequest {
+		p := &GetScheduleRequest{
+			Schedules:                schedules, // need to chunk these out
```
levb: > need to chunk these out

elaborate?

server/remote/schedule.go
```diff
@@ -0,0 +1,22 @@
...
+	// 0= free, 1= tentative, 2= busy, 3= out of office, 4= working elsewhere.
+	// example "0010", which means free for first and second block, tentative for third, and free for fourth
+	AvailabilityView string `json:"availabilityView,omitempty"`
+
+	// ScheduleItems []interface{} `json:"scheduleItems,omitempty"`
+	// WorkingHours interface{} `json:"workingHours,omitempty"`
+	// Error *FreeBusyError `json:"error,omitempty"`
+}
+
+type GetScheduleResponse struct {
```
levb: 0/5 prefer embedding the `{ Value []*sometype }` in the `msgraph` functions themselves, no need exposing them as generic `remote` structs IMO.

server/store/user_store.go
```diff
@@ -67,6 +83,38 @@ func (s *pluginStore) StoreUser(user *User) error {
 		return err
 	}
 
+	var allUsers []*UserShort
```
levb: 0/5 aggregating multiple keys should be all done in the `api` layer. The `store` layer should define the persistent structs, and provide (atomic/simple) operations on them, the lowest common denominator; abstracting away only things like key hashing, encryption, simple collision handling, etc.

See https://github.com/levb/mattermost-plugin-solar-lottery/blob/master/server/store/rotation_store.go#L11 for an example.

server/api/availability.go
mickmister: Correct, it is graph side. It is the only option as the unique identifier for the user when using the `getSchedule` API.

server/remote/schedule.go
```diff
@@ -0,0 +1,22 @@
...
+// See License for license information.
+
+package remote
+
+// ScheduleInformation undocumented
+type ScheduleInformation struct {
+	// Email of user
+	ScheduleID string `json:"scheduleId,omitempty"`
+
+	// 0= free, 1= tentative, 2= busy, 3= out of office, 4= working elsewhere.
```
crspeller: This comment seems to be inaccurate.

server/remote/schedule.go
```diff
@@ -0,0 +1,22 @@
...
+	// example "0010", which means free for first and second block, tentative for third, and free for fourth
+	AvailabilityView string `json:"availabilityView,omitempty"`
```
crspeller: 0/5 Maybe define a type for this? `type Availability string`

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
+package msgraph
+
+import (
+	"strconv"
+
+	"github.com/mattermost/mattermost-plugin-msoffice/server/remote"
+)
+
+type GetScheduleSingleResponse struct {
```
crspeller: Any need to export these?

server/remote/msgraph/get_schedule_batched.go
```diff
@@ -0,0 +1,104 @@
...
+
+	//  This is where we can simulate large batches
+	// 	TODO: Split up emails given into different batches properly
+	allRequests := []*SingleRequest{}
+	// allRequests = append(allRequests, makeRequest())
+
+	numRequestsInBatch := 1
+	// numRequestsInBatch := 20
+
+	for i := 0; i < numRequestsInBatch; i++ {
```
crspeller: Is this for testing? I am not sure what this function is trying to accomplish.

server/remote/schedule.go
mickmister: The availabilities are documented [here](https://docs.microsoft.com/en-us/graph/outlook-get-free-busy-schedule) about halfway down the page

server/remote/msgraph/get_schedule_batched.go
mickmister: Right, these should not be exported

server/remote/msgraph/get_schedule_batched.go
mickmister: Yes, this was a quick and dirty way to test the sizes of batches the MSGraph API accepts, because it is not documented. This will be cleaned up in this PR.

server/remote/schedule.go
levb: types aliases are in fashion these days

server/api/availability.go
mickmister: https://mattermost.atlassian.net/browse/MM-21649

Removed cron dependency in favor of using a ticker select loop, so we have more control.

server/api/api.go
mickmister: I see you've marked this as resolved. Do you not want it addressed in this PR?

mickmister:
This is ready for review. I've created tickets to track further work in the project's epic https://mattermost.atlassian.net/browse/MM-19881

server/remote/msgraph/call.go
```diff
@@ -11,12 +11,29 @@ import (
...
+	contentType := "application/json"
+	buf := &bytes.Buffer{}
+	err = json.NewEncoder(buf).Encode(in)
+	if err != nil {
+		return nil, err
+	}
+	return c.Call(method, path, contentType, buf, out)
+}
+
+func (c *client) CallURLEncodedForm(method, path string, in url.Values, out interface{}) (responseData []byte, err error) {
```
levb: 0/5
1. Maybe just `CallFormPost` or even `CallForm`?
2. With the addition of these 2 methods, maybe hide `Call` as private? Seems redundant now.

server/remote/msgraph/get_notification_data.go
```diff
@@ -17,7 +17,7 @@ func (c *client) GetNotificationData(orig *remote.Notification) (*remote.Notific
 	switch wh.ResourceData.DataType {
 	case "#Microsoft.Graph.Event":
 		event := remote.Event{}
-		_, err := c.Call(http.MethodGet, wh.Resource, nil, &event)
+		_, err := c.CallJSON(http.MethodGet, wh.Resource, nil, &event)
```
levb: 👍

server/remote/client.go
```diff
@@ -3,11 +3,17 @@
...
+import (
+	"io"
+	"net/url"
+	"time"
+)
 
 type Client interface {
 	AcceptUserEvent(userID, eventID string) error
-	Call(method, path string, in, out interface{}) (responseData []byte, err error)
+	Call(method, path, contentType string, in io.Reader, out interface{}) (responseData []byte, err error)
```
levb: see my comment below, maybe remove `Call` as redundant? (This would also ensure that we always set the right content type in the calls.

server/plugin/plugin.go
```diff
@@ -67,10 +68,13 @@ func (p *Plugin) OnActivate() error {
 	}
 
 	p.httpHandler = http.NewHandler()
+
 	p.notificationHandler = api.NewNotificationHandler(p.newAPIConfig())
 
 	command.Register(p.API.RegisterCommand)
 
+	p.initUserStatusSyncJob()
```
levb: I wonder if it's needed here, or OnConfigurationChange is enough.

mickmister:
@levb I've addressed the your feedback in the latest commit.

server/store/user_store.go
```diff
@@ -85,3 +135,33 @@ func (s *pluginStore) DeleteUser(mattermostUserID string) error {
...
+	result := map[string]*UserShort{}
+
+	for _, u := range index {
+		result[u.RemoteID] = u
+	}
+
+	return result
+}
+
+func (index UserIndex) ByEmail() map[string]*UserShort {
```
mickmister: Instead of deserializing into a map, I have chosen to have these helper methods that create maps for the different use cases. The `ByEmail` method is the only one used at the moment. The email is unique per each Microsoft account.

server/api/status_sync_job.go
```diff
@@ -0,0 +1,71 @@
...
+	"sync"
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
+)
+
+const JOB_INTERVAL = 5 * time.Minute
+
+type StatusSyncJob struct {
+	api        API
```
crspeller: This is confusing because we usually call the plugin API this.

server/plugin/plugin.go
```diff
@@ -207,3 +213,26 @@ func (p *Plugin) loadTemplates(bundlePath string) error {
 	p.Templates = templates
 	return nil
 }
+
+func (p *Plugin) initUserStatusSyncJob() {
```
crspeller: This is going to get run once for every node. Can we put a comment here to make sure that is known?

server/api/status_sync_job.go
levb: I think that once the other is renamed to `PluginAPI` and `p` is not replicated in the code, `api` becomes self-explanatory; but I'd love a better name; maybe `app` (for everything, package and all) despite all of its negative properties?

server/plugin/plugin.go
levb: I started naming methods that are not ready for production more explicitly, `Debug...` for instance (or maybe POC in this case?). 0/5 Intention-revealing names are better than comments.

server/api/availability.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	if len(s.AvailabilityView) == 0 {
+		return "No availabilities found", nil
+	}
+
+	av := s.AvailabilityView[0]
+	return api.setUserStatusFromAvailability(api.mattermostUserID, status.Status, av), nil
+}
+
+func (api *api) SyncStatusForAllUsers() (string, error) {
```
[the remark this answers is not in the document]
mickmister: I've made the `SyncStatusForUsers` method and have the other two methods calling that one, to reduce code duplication.

server/api/status_sync_job.go
mickmister: Separate ticket for renaming package if we plan on doing so?

mickmister:
@levb @crspeller Please re-review when you have the chance

mickmister:
@crspeller So far, we have not gotten @DHaussermann involved with testing. I'll meet with him and get him set up to test master.

server/api/availability.go
mickmister: @levb If I understand correctly, the reason we may want to map it to a user is so we can always have a user id to use for routes that require it, unless there is also another reason (code simplification?). Would we have the admins create a user in Azure dedicated to the plugin?

Ideally we use the application-level permissions as they are designed for the use case of a bot, as opposed to user-delegated permissions. The application-level permissions work a bit differently and have more of a blanket of access to different operations. However, we can map to a dedicated user /and/ use application-level permissions for the operations.

PR #19, opened 2019-12-19
[MM-19897] Improvements for `/msoffice createevent` and `/msoffice findmeetings`

jfrerich:
#### Summary
This PR adds enhancements for POC `/msoffice createevent`, but also enhances the CLI flag options and fixes the `FindMeetings()` API method so it works.

#### CreateEvent
 * get MM_UserIDs from CLI and accept them in CreateEvent() to check if mapped to msoffice account
 * implement CLI flags for the following:
   * help 
   * subject
   * body
   * allday
   * reminder
   * location attendees

#### FindMeetings
  * move find meetings structs to its own file
  * correct MeetingTimeSuggestionsResults struct fields to correctly unmarshal
  * properly implement function from previous POC

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19897

#### TODO 

The following issues need to be resolved.  Creating individual tickets for each.

- [x] `command/create_event.go` - allow flags that accept spaces (subject, body)
- [x] `command/create_event.go` - improve `--location` flag to include separate location types and better argument parsing
- [x] `command/create_event.go` - if  `allday`  flag set, make sure user provides start and end times and that times are different days and at midnight
- [x] `msgraph/find_meeting_times.go` - check `meetingsOut.EmptySuggestionReason`  field for non-nil value and report err.  Possible situation is organizer is busy when trying to setup a meeting
- [x] `api/calendar.go`  - when checking MM userID values, if the user is mapped, add to the `calendarEvent` struct 
- [x] write tests for API layer


server/plugin/command/create_event.go
```diff
@@ -12,10 +13,15 @@ import (
 
 func getCreateEventFlagSet() *flag.FlagSet {
 	flagSet := flag.NewFlagSet("create", flag.ContinueOnError)
-	flagSet.String("subject", "", "Subject of the Event (no spaces for now)")
+	flagSet.Bool("help", false, "show help")
+	flagSet.String("subject", "", "Subject of the event (no spaces for now)")
```
levb: Since subject and body are not practical to pass in as arguments, I recommend recommend naming these flags `--test-subject` etc. to be more intention revealing. (also `...body, location`)

server/plugin/command/create_event.go
```diff
@@ -12,10 +13,15 @@ import (
...
+	flagSet.String("body", "", "Body of the event (no spaces for now)")
+	flagSet.StringSlice("location", []string{}, "Location of the event <displayName,street,city,state,postalcode,country> (comma separated; no spaces)")
 	flagSet.String("starttime", time.Now().Format(time.RFC3339), "Start time for the event")
+	flagSet.Bool("allday", false, "Set as all day event (starttime/endtime must be set to midnight on different days - 2019-12-19T00:00:00-00:00)")
+	flagSet.Int32("reminder", 15, "Reminder (in minutes)")
 	flagSet.String("endtime", time.Now().Add(time.Hour).Format(time.RFC3339), "End time for the event")
-	flagSet.StringSlice("attendees", []string{}, "A comma separated list of Attendees")
+	flagSet.StringSlice("attendees", []string{}, "A comma separated list of Mattermost UserIDs")
```
levb: 1/5 I recommend accepting `@username`s, see https://github.com/levb/mattermost-plugin-solar-lottery/blob/a37a0ba0512bad58ced4a542740ec7cf9b53aa3c/server/plugin/api.go#L28 for an example.

server/plugin/command/create_event.go
```diff
@@ -26,121 +32,100 @@ type userError struct {
 
 func parseCreateArgs(args []string) (*remote.Event, *userError, error) {
 
-	var attendees []*remote.Attendee
```
levb: 👍

server/plugin/command/create_event.go
```diff
@@ -12,10 +13,15 @@ import (
 
 func getCreateEventFlagSet() *flag.FlagSet {
 	flagSet := flag.NewFlagSet("create", flag.ContinueOnError)
-	flagSet.String("subject", "", "Subject of the Event (no spaces for now)")
+	flagSet.Bool("help", false, "show help")
```
levb: 0/5 we should use the same style consistently everywhere, I believe other plugins usually use the `...Var` form of the functions.

server/plugin/command/create_event.go
```diff
@@ -26,121 +32,100 @@ type userError struct {
...
+		return nil, &userError{ErrorMessage: fmt.Sprintf(getCreateEventFlagSet().FlagUsages())}, nil
+	}
+
 	subject, err := createFlagSet.GetString("subject")
 	if err != nil {
 		return event, nil, err
 	}
 	// check that next arg is not a flag "--"
-	if subject == "" || strings.HasPrefix(subject, "--") {
+	if strings.HasPrefix(subject, "--") {
```
levb: ??? I don't understand this check. The default value appears to be "", not "--".

server/plugin/command/create_event.go
```diff
@@ -26,121 +32,100 @@ type userError struct {
...
 
-	//                //
-	// Optional Flags //
-	//                //
+	body, err := createFlagSet.GetString("body")
+	if err != nil {
+		return event, nil, err
+	}
+	// check that next arg is not a flag "--"
+	if strings.HasPrefix(body, "--") {
```
levb: ditto

server/plugin/command/create_event.go
```diff
@@ -12,10 +13,15 @@ import (
...
 func getCreateEventFlagSet() *flag.FlagSet {
 	flagSet := flag.NewFlagSet("create", flag.ContinueOnError)
-	flagSet.String("subject", "", "Subject of the Event (no spaces for now)")
+	flagSet.Bool("help", false, "show help")
+	flagSet.String("subject", "", "Subject of the event (no spaces for now)")
+	flagSet.String("body", "", "Body of the event (no spaces for now)")
+	flagSet.StringSlice("location", []string{}, "Location of the event <displayName,street,city,state,postalcode,country> (comma separated; no spaces)")
 	flagSet.String("starttime", time.Now().Format(time.RFC3339), "Start time for the event")
+	flagSet.Bool("allday", false, "Set as all day event (starttime/endtime must be set to midnight on different days - 2019-12-19T00:00:00-00:00)")
+	flagSet.Int32("reminder", 15, "Reminder (in minutes)")
```
levb: Why `Int32` and not `Int`?

server/plugin/command/create_event.go
```diff
@@ -26,121 +32,100 @@ type userError struct {
...
 	if err != nil {
 		return event, nil, err
 	}
 	if strings.HasPrefix(endTime, "--") {
-		return event, &userError{ErrorMessage: "must specify an event subject"}, nil
+		return event, &userError{ErrorMessage: "must specify an endtime"}, nil
+	}
+	event.End = &remote.DateTime{
+		DateTime: endTime,
+		TimeZone: "Pacific Standard Time",
```
levb: This should be extracted from t.Location information. Please file a separate ticket to handle the user time zone properly - the meeting should be created in the local timezone.

server/plugin/command/create_event.go
```diff
@@ -26,121 +32,100 @@ type userError struct {
...
+	reminder, err := createFlagSet.GetInt32("reminder")
+	if err != nil {
+		return event, nil, err
+	}
+	if strings.HasPrefix(strconv.Itoa(int(reminder)), "--") {
+		return event, &userError{ErrorMessage: "must specify an reminder"}, nil
+	}
+	event.ReminderMinutesBeforeStart = reminder
+
+	location, err := createFlagSet.GetStringSlice("location")
```
levb: Thinking forward here. Gathering 6 parameters (or even trying to pre-fill them using what we know about the user) seems impractical from the UX standpoint. This is begging for an intelligent autocomplete. Is there a Microsoft Graph API we can use for this? Please file a separate Jira ticket on how we should support Location when creating events.

Of particular importance are video calls. There may be some overlap with Location since these are interrelated datum in the UX.

server/plugin/command/find_meeting_times.go
```diff
@@ -4,28 +4,37 @@
...
+	"strings"
 
 	"github.com/mattermost/mattermost-plugin-msoffice/server/remote"
 	"github.com/mattermost/mattermost-plugin-msoffice/server/utils"
 )
 
 func (c *Command) findMeetings(parameters ...string) (string, error) {
 	meetingParams := &remote.FindMeetingTimesParameters{}
-	// meetings, err := c.API.FindMeetingTimes(time.Now(), time.Now().Add(14*24*time.Hour))
+
```
levb: 1/5 We should support entering users as `@username`, or as a `myaddress@email.sucks`, see above comment for an example.

1/5 recommend a consistent syntax for providing a list of users, 0/5 suggest `--users @user1,@user2`. Using a space-separated list (i.e. the tail of the command line) is appealing, but then it is also appealing to use for other purposes, like typing in a subject. Ultimately, we may chose something in between consistency and practicality, e.g. going for `/... invite @user1 @user2 2020-01-01 at 12:00AM Celebrate New Year's ` and smart parsing based on the values themselves.

server/remote/meeting.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See License for license information.
+
+package remote
+
+import "time"
+
+type AttendeeBase struct {
```
levb: I don't understand why we need 2 Attendee types - what is the purpose of having a separate struct and why can it not be accomplished with either omitempty or a custom to/from json marshal functions.

server/plugin/command/create_event.go
jfrerich: That's fine.  I'll prepend `test-` to the `location`, `body`, `subject` flags.  Note that I have added a jira ticket to address the `no spaces` issue.

https://mattermost.atlassian.net/browse/MM-21347

jfrerich: The `location` flag does work.  I considered the tradeoff between requiring one flag `location` with the comma separated list of values, versus splitting `location` into several flags.  We could introduce `--location.street`, `--location.city`, `--location.state`, .etc as a alternative method.  the tradeoff being that a user would be required to list many extra flags.

Ticket created:
https://mattermost.atlassian.net/browse/MM-21390

server/plugin/command/create_event.go
jfrerich: Agreed. I filed a separate ticket.

https://mattermost.atlassian.net/browse/MM-21391

server/plugin/command/create_event.go
jfrerich: Int32 was specified in the MS Graph API and I was just keeping the type consistent.  
image

server/plugin/command/create_event.go
jfrerich: Ticket filed here: 
https://mattermost.atlassian.net/browse/MM-21391

server/remote/meeting.go
jfrerich: I will remove this struct.  With the `omitempty` setting inside the `Attendee` struct, we can share only one `Attendee` type.

server/plugin/command/create_event.go
jfrerich: @levb, can you add some context to this comment?  Are you suggesting some sort of variadic function usage?

levb: No, see `flagSet.StringVar` etc. It's just a slightly different way of using the FlagSet. Not even a better one, just advocating that we do the same everywhere.

server/api/calendar.go
```diff
@@ -27,7 +27,19 @@ func (api *api) CreateCalendar(calendar *remote.Calendar) (*remote.Calendar, err
 	return client.CreateCalendar(calendar)
 }
 
-func (api *api) CreateEvent(calendarEvent *remote.Event) (*remote.Event, error) {
+func (api *api) CreateEvent(calendarEvent *remote.Event, IDs []string) (*remote.Event, error) {
```
mickmister:
```suggestion
func (api *api) CreateEvent(calendarEvent *remote.Event, userIDs []string) (*remote.Event, error) {
```

server/plugin/command/create_event.go
```diff
@@ -12,10 +13,15 @@ import (
 
 func getCreateEventFlagSet() *flag.FlagSet {
 	flagSet := flag.NewFlagSet("create", flag.ContinueOnError)
-	flagSet.String("subject", "", "Subject of the Event (no spaces for now)")
+	flagSet.Bool("help", false, "show help")
+	flagSet.String("test-subject", "", "Subject of the event (no spaces for now)")
+	flagSet.String("test-body", "", "Body of the event (no spaces for now)")
```
mickmister: Can flagSet handle spaces for you, or is it considered the "end" of the flag when a space is reached?

server/plugin/command/find_meeting_times.go
mickmister: > Ultimately, we may chose something in between consistency and practicality, e.g. going for /... invite @user1 @user2 2020-01-01 at 12:00AM Celebrate New Year's and smart parsing based on the values themselves.

I think this is very tedious UX. I certainly would not want to manually type out dates or times. Ideally we would have an as-simple-as-possible equivalent in a UI form. I think it would be about 8 mana to implement, without username autocomplete for the attendees. Everything else would be entered by value as far as I can tell. Datetime components are pretty straightforward to configure.

server/plugin/command/create_event.go
jfrerich: I did a quick search through the plugins I have cloned it looks like most use the non-`Var` form of the functions.  Specifically, I was looking at the cloud plugin as an example. 

Perhaps we need to actually we can discuss this in our meeting?  I assume there will be arguments for both methods.

server/plugin/command/create_event.go
levb: At first glance it (pflag) appeared very customizable and transparent to me, but I did not dig in. It does seem a little suspect that it takes a `params ...string` rather than a `command string` as its input.

server/api/calendar.go
levb: I'd recommend going as far as establishing a convention of a `MattermostUserID` vs `RemoteUserID` and complying with it throughout. Names should be intention revealing, but most importantly consistent - same name should mean the same thing. Practical exceptions are often made for field names, but for variables like this I don't think it's a good excuse. Code should read like a book - clear, but not much redundancy :)

On that note why not s/calendarEvent/event/? What is `calendar` referring to here that is not already implied by `event`?

jfrerich: I prefer `MattermostUserID` for this name because there are several `IDs` (calendar, MM UserId, Outlook User ID)

Also agree with `s/calendarEvent/event/` as `event` does imply a calendar event.

server/plugin/command/create_event.go
jfrerich: During the POC phase, I couldn't get find a method to have flagSet handle the spaces.  My intention was to create a separate ticket for this issue, but could try and get it into this PR.

server/plugin/command/create_event.go
jfrerich: There appears to be a few limitations of the go `flag` and the `pflags` libraries.  

1. The library expects a non-bool flag to have a value and does not discriminate if the value is the next flag.  
Example `--flag --nextflag`, the value of the `--flag` is going to be `--nextflag`.  This `strings.HasPrefix(subject, "--")` checks that case and returns an error when it occurs.

However, given the example `--flag flagvalue --nextflag`, where `--nextflag` is end-of-line, the parser will let the user know there is no value for the `--nextflag` flag

2. There is know required setting for flag.  We have to define which flags are required and handle to code to ensure the flags are set.  I have this code in the following commit.

server/plugin/command/create_event.go
jfrerich: see above comment

server/plugin/command/create_event.go
jfrerich: @levb, we need to continue the discussion on accessing the plugin API for this to happen.

server/plugin/command/create_event.go
levb: Leave as is, it's fine.

server/plugin/command/create_event.go
levb: Right, got it. 0/5 I'd have made a different choice to keep the implementation simple and clearly intention-revealing, but I understand the considerations now. 👍

server/plugin/command/create_event.go
```diff
@@ -2,147 +2,142 @@ package command
...
-	flagSet.StringSlice("attendees", []string{}, "A comma separated list of Attendees")
+	flagSet.StringSlice("attendees", nil, "A comma separated list of Mattermost UserIDs")
 
 	return flagSet
 }
 
-type userError struct {
-	ErrorMessage string
-}
+func parseCreateArgs(args []string) (*remote.Event, error) {
```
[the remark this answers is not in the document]
jfrerich: done.

server/api/calendar.go
```diff
@@ -27,13 +27,25 @@ func (api *api) CreateCalendar(calendar *remote.Calendar) (*remote.Calendar, err
 	return client.CreateCalendar(calendar)
 }
 
-func (api *api) CreateEvent(calendarEvent *remote.Event) (*remote.Event, error) {
+func (api *api) CreateEvent(event *remote.Event, MattermostUserIDs []string) (*remote.Event, error) {
```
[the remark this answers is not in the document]
jfrerich: yep. Done.

server/api/api.go
```diff
@@ -29,7 +29,7 @@ type Subscriptions interface {
 
 type Calendar interface {
 	ViewCalendar(from, to time.Time) ([]*remote.Event, error)
-	CreateEvent(event *remote.Event) (*remote.Event, error)
+	CreateEvent(event *remote.Event, userIDs []string) (*remote.Event, error)
```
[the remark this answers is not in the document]
jfrerich: done.

server/plugin/command/create_event.go
jfrerich: added to https://mattermost.atlassian.net/browse/MM-21390

PR #25, opened 2020-01-24
[MM-21689] Handle timezone complications between Go and Microsoft Graph

mickmister:
#### Summary

This PR handles two things regarding timezones:
- The conversion of Microsoft-compatible datetime values to/from Go-compatible datetime values.
- When processing user-provided data, or displaying data to the user, the user's calendar timezone is taken into account. This is applied to the `viewcal` and `findmeetings`, and `createevent` commands at the moment.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21689


server/utils/timezones.go
```diff
@@ -0,0 +1,608 @@
...
+}
+
+func populateConversionMap() {
+	for _, entry := range allTimezones {
+		windowsToIanaMap[entry[1]] = entry[0]
+	}
+}
+
+var allTimezones = [][]string{
+	{"Africa/Abidjan", "Greenwich Standard Time", "+00:00"},
```
mickmister: This method of timezone name conversion is not ideal, but I have not found any conversions between Go-compliant and MS Graph-compliant datetimes available to use.

Source of the data in the array:
https://techsupport.osisoft.com/Documentation/PI-Web-API/help/topics/timezones/iana.html

server/remote/common.go
```diff
@@ -44,3 +80,12 @@ func (dt DateTime) Time() time.Time {
 	}
 	return t
 }
+
+func safeTimeZone(timeZone string) string {
```
mickmister: Not sure on the name here. This function takes a time zone, and converts it to a Go-compatible one if it is not compatible.

server/remote/msgraph/get_user_mailbox_settings.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+)
+
+func (c *client) GetUserMailboxSettings(remoteUserID string) (*remote.MailboxSettings, error) {
+	var u string
+	if remoteUserID == "me" {
```
levb: Do we need this logic? 0/5 with the exception of the first contact(?) are we ever going to use /me? We always have the UserID available to us, do we not?

server/remote/common.go
levb: well, why not call it `GoTimeZone` (and `MicrosoftTimeZone`)? Also I'd make it public, if nothing else for documentation purposes so they appear in GoDoc.

server/utils/timezones.go
```diff
@@ -0,0 +1,610 @@
+package utils
```
levb: 1. It's really hard to believe that this mapping does not yet exist in goland as it appears to be a common use case interacting with Microsoft software in general.
2. If it does not we 0/5 should work on making our implementation Google-able (separate repo? a submission to one of the MSGraph go clients?) in hope to make it used outside of mattermost.
3. Either way, suggest a sub-package under utils for the time being, `utils/tz`?

server/utils/timezones.go
```diff
@@ -0,0 +1,610 @@
...
+
+var windowsToIanaMap = map[string]string{}
```
levb: nit: suggest `s/windowsToIanaMap/windowsToIANA/

server/utils/timezones.go
```diff
@@ -0,0 +1,610 @@
...
+
+func ConvertWindowsTimezoneToIANA(winTZ string) string {
+	if len(windowsToIanaMap) == 0 {
```
levb: `init()`

server/utils/timezones.go
levb: 1/5 file a HW ticket to have a `make tz` that would pull the list from the link ^^ and generate the go map. This list is not as constant as one would like. Bonus: have a hardcoded default version, and pull the current one dynamically on startup.

server/utils/timezones.go
```diff
@@ -0,0 +1,610 @@
...
+
+	tz, ok := windowsToIanaMap[winTZ]
+	if ok {
+		return tz
+	}
+
+	return ""
+}
+
+func populateConversionMap() {
```
levb: Seems unnecessary since a few substitutions can make this into an idiomatic go map.

levb:
overall comment on naming the util functions, `tz.Go()`, and `tz.Microsoft()` seem to be the most intention-revealing, concise names?

server/api/calendar.go
```diff
@@ -76,3 +76,28 @@ func (api *api) GetUserCalendars(userID string) ([]*remote.Calendar, error) {
...
 	return client.GetUserCalendars(me.ID)
 }
+
+func (api *api) GetUserTimezone(mattermostUserID string) (string, error) {
+	client, err := api.MakeClient()
+	if err != nil {
+		return "", err
+	}
+
+	var remoteUserID string
```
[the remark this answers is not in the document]
mickmister: @crspeller Can you confirm I've addressed this in the most recent push?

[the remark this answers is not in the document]
mickmister: I'm actually not sure what the ask is here.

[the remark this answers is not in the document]
crspeller: Sorry I wasn't clear. I like the split out you did. I really meant the `withX` stuff the solar lottery plugin does.

server/utils/timezones.go
mickmister: https://golang.org/src/time/zoneinfo_abbrs_windows.go

The one resource I've found pertaining to using Windows timezones in the Go library is a non-exported map from the `time` package. Any use of the map seems to be internal to the library. The keys are Microsoft-compliant, and the values are abbreviations of timezones. These abbreviations are not compliant with `time.LoadLocation` across the board.

There is a comment to the right of each entry, showing an example IANA format for each Windows format. This is a much better resource, since it's directly from Go. Each of the timezones are compatible with `time`, as shown in the unit test I've included. I dismissed this originally as the information is in the comments, but we can parse the comments in the `make tz` script.

server/remote/common.go
```diff
@@ -3,7 +3,11 @@
 
 package remote
 
-import "time"
+import (
+	"time"
+
+	tz "github.com/mattermost/mattermost-plugin-mscalendar/server/utils/tz"
```
levb: why the alias?

server/utils/tz/conversion.go
```diff
@@ -0,0 +1,743 @@
...
+	"Hawaiian Standard Time":          "Pacific/Honolulu",
+	"Line Islands Standard Time":      "Pacific/Kiritimati",
+	"Marquesas Standard Time":         "Pacific/Marquesas",
+	"Norfolk Standard Time":           "Pacific/Norfolk",
+	"West Pacific Standard Time":      "Pacific/Port_Moresby",
+	"Tonga Standard Time":             "Pacific/Tongatapu",
+}
+
+// Source: https://techsupport.osisoft.com/Documentation/PI-Web-API/help/topics/timezones/iana.html
+var iANAToWindows = map[string]string{
```
levb: nit: if it's an exact reverse, you can in this case build it from the first map, in `init()`? Also, 1/5 the map definition should be in its own file, separate from the code. Suggest, 0/5 `data.go`.

server/remote/msgraph/get_user_mailbox_settings.go
```diff
@@ -0,0 +1,18 @@
...
+package msgraph
+
+import (
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+)
+
+func (c *client) GetUserMailboxSettings(remoteUserID string) (*remote.MailboxSettings, error) {
+	u := "/users/" + remoteUserID + "/mailboxSettings"
```
levb: Is there a URL builder available for this in the graph client?

server/remote/common.go
```diff
@@ -32,8 +38,37 @@ func (dt DateTime) String() string {
...
 
+func (dt DateTime) PrettyString() string {
+	t := dt.Time()
+	if t.IsZero() {
+		return "n/a"
+	}
+	return t.Format(time.RFC822)
+}
+
+func (dt DateTime) ConvertToTimezone(timeZone string) *DateTime {
```
levb: 0/5 `s/ConvertToTimezone/InTimezone/

server/remote/common.go
```diff
@@ -32,8 +38,37 @@ func (dt DateTime) String() string {
...
+	t := dt.Time()
+	if t.IsZero() {
+		return &dt
+	}
+
+	tz := tz.Go(timeZone)
```
levb: 0/5 inline?

server/remote/common.go
```diff
@@ -16,11 +20,13 @@ type DateTime struct {
 
 const RFC3339NanoNoTimezone = "2006-01-02T15:04:05.999999999"
 
-func NewDateTime(t time.Time) *DateTime {
+// NewMicrosoftDateTime creates a DateTime that is compatible with Microsoft's API.
+func NewMicrosoftDateTime(t time.Time, timeZone string) *DateTime {
```
levb: _really_ not sure, but in the caller context, woudn't `remote.NewDateTime()` be sufficiently intention-revealing?

server/remote/common.go
```diff
@@ -3,7 +3,11 @@
 
 package remote
 
-import "time"
+import (
+	"time"
```
levb: ./remote/date_time.go?

server/plugin/command/view_calendar.go
```diff
@@ -15,8 +15,23 @@ func (c *Command) viewCalendar(parameters ...string) (string, error) {
...
 
+	var timeZone string
+	tz, err := c.API.GetUserTimezone(c.Args.UserId)
+	if err == nil {
+		timeZone = tz
+	}
+
+	if timeZone != "" {
+		for _, event := range events {
+			event.Start = event.Start.ConvertToTimezone(timeZone)
```
levb: y, I think `event.Start.InTimezone(timeZone)`, or even better `event.Start.In(timeZone)` would be typographically better, and sufficiently intention-revealing. (see comments below, reading from the bottom)

levb: `ConvertToTimezone` is too Java, not poetic enough :)  (poetic == reduced, simplified)

server/api/event.go
```diff
@@ -3,7 +3,31 @@
...
 package api
 
-import "github.com/pkg/errors"
+import (
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+)
+
+func (api *api) CreateEvent(event *remote.Event, mattermostUserIDs []string) (*remote.Event, error) {
```
levb: 👍 thanks for the cleanup!

levb:
Totally OT, 0/5: I really like looking at the godoc for each package I make, should do it more. Without much actual contents of comments, it should be readable, intuitive, and typographically sound. Combine that with never indenting the go code over 4 deep, and one is 80% perfect.

server/utils/tz/conversion.go
mickmister: It's not an exact reverse. Windows -> IANA is one-to-many. I'd like to discuss the implications with you when we can get together.

server/api/calendar.go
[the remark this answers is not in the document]
levb: Coming in https://github.com/mattermost/mattermost-plugin-mscalendar/pull/29

PR #26, opened 2020-01-24
[MM-21691] Handle errors for getSchedule call

mickmister:
#### Summary

This PR takes into account when a Graph API error occurs with an individual call inside of the batch call for `getSchedule`. If there is an error with any of them, the operation bails and no user statuses get updated. The way the caller is set up, it should be fairly trivial to change the code to simply "skip over" the ones that had errors if we wish to do it that way.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21691

### DIFF
--- a/server/remote/msgraph/get_schedule_batched.go
+++ b/server/remote/msgraph/get_schedule_batched.go
@@ -1,6 +1,7 @@
 package msgraph
 
 import (
+	"errors"
 	"strconv"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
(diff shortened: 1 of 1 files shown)


jfrerich:
@mickmister semi-unrelated to this ticket, should we `s/schedules/scheduleIDs/`?  A schedule is a string type and schedules makes it feel like I'd expect a struct with fields, instead of any array of strings.

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/master/server/remote/msgraph/get_schedule_batched.go#L28

mickmister:
@jfrerich Microsoft usually refers to them as schedules which is why it's here. But I agree, scheduleIDs is a better choice :+1:

Issue #24, opened 2020-01-23
Implement user `disconnect` slash command


mickmister:
#### Summary

The user should be able to disconnect their Mattermost account from their Microsoft account at any time. Using a `/mscalendar disconnect` command, any user data in the plugin's kv store should be cleaned up.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21991

hanzei:
@mickmister Should this be a HW?

mickmister:
@hanzei Yes I had meant to set up the titles/labels as such. Thanks for catching this!

PR #32, opened 2020-02-05
[MM-21688] Use dedicated bot account in Azure

mickmister:
#### Summary

We are using the `getSchedule` Graph API endpoint to fetch user availabilities, and using application-level permissions to do so. The request requires us to supply a user ID, so we need a dedicated user in Azure to reference.

This PR allows the MM admin to connect the mscalendar bot to an account in Azure, using the `/mscalendar connect_bot` command.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21688

#### Additional Info

Before running `/mscalendar connect_bot`, `/mscalendar availability` should fail, since we do not have a bot user to use for the user id.

Todo:

- [x] `mscalendar/oauth2.go` needs to have access to `IsAuthorizedAdmin` so we can ensure only admins can connect the bot account
	- `IsAuthorizedAdmin` is now `bot.IsUserAdmin` and `PluginAPI.IsSysAdmin`. Shouldn't we want to union these two, since a sys admin should be treated as a plugin admin?
- [x] fix `command/connect_test.go` and `command/disconnect_test.go` once ^ has been resolved
- [x] `oauth2_connect_test.go` and `oauth2_complete_test.go` have been deleted, but there are some tests I have written to assert things pertaining to connecting the bot account

- [ ] Write README instructions for creating dedicated account in Azure

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/24


jfrerich:
@mickmister, is this PR ready for review?  I usually clone the repo to make checking some of the code easier to traverse and some of the methods don't exist.

server/command/connect.go
```diff
@@ -7,11 +7,34 @@ import (
 	"fmt"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/config"
+	"github.com/pkg/errors"
 )
 
 func (c *Command) connect(parameters ...string) (string, error) {
+	u, err := c.API.GetRemoteUser(c.Args.UserId) // needs fix
```
levb: Maybe not in this PR, but we should clean out `plugin.API` references and use `mscalendar.PluginAPI` instead.

server/mscalendar/availability.go
```diff
@@ -35,11 +35,25 @@ func (m *mscalendar) SyncStatusAll() (string, error) {
 		return "", err
 	}
 
-	mmIDs := userIndex.GetMattermostUserIDs()
+	index := userIndex.GetMattermostUserIDs()
+	mmIDs := []string{}
+	for _, id := range index {
+		if id != m.Config.BotUserID {
```
levb: Is this the best place to do filtering? Maybe at the user-API level is better? Also, why would the Bot user ID be in the list to start with, out of curiosity?

server/remote/remote.go
```diff
@@ -15,7 +15,7 @@ import (
 
 type Remote interface {
 	MakeClient(context.Context, *oauth2.Token) Client
-	MakeSuperuserClient(context.Context) Client
+	MakeSuperuserClient(context.Context, string) Client
```
levb: give the parameters names, to reveal the intention of the `string` parameter?

mickmister:
@jfrerich I should have marked the PR as draft since it is not properly building at the moment. `IsAuthorizedAdmin` in particular does not exist anymore after the project restructure, and needs to be exposed properly to some of the `oauth` related files. Is this the function you are referring to? The `connect.go`, `disconnect.go`, and `oauth*` files have these issues. There are traces of `GetRemoteUser` as well, as @levb as pointed out.

@levb I need to discuss with you how to expose this admin-checking functionality to the `oauth` package, as only trusted admins should be able to connect/disconnect the bot from msgraph. The `oauth.App` interface is more or less isolated from the rest of the packages at the moment (on purpose), but needs to have access to these admin-checking functions, as noted in a todo in the PR description. Or we need to restructure the flow to allow for checking before the `oauth` package receives the request.

server/command/connect.go
mickmister: I mistakenly left this here. It should be using `withUserExpanded` `ExpandUser`, or `LoadUser`.

server/mscalendar/availability.go
mickmister: The bot is inside of the index because it is placed there when the oauth completes. We can/should avoid placing the bot's account in the index.

mickmister:
@jfrerich I can comment out certain things that need to be restructured/added so it will properly compile, though the tests will still fail without the added functions.

server/mscalendar/availability.go
levb: Right. I like the precaution - letting it through would be wasteful if harmless.

mickmister: @levb

> Maybe at the user-API level is better?

I don't see how that is an option, as this is for the status sync job

>  We can/should avoid placing the bot's account in the index.

The `UserStore` does not have access to the config (which i agree with that decision), but this means it does not know what the bot's user id is. I'm thinking the store should have a separate method for storing a user in the index, but then the caller of `StoreUser()` will always need to know to call `StoreUserIndex()` as well.

mickmister:
@jfrerich This PR is ready for review now

server/utils/pluginapi/api.go
```diff
@@ -26,14 +26,19 @@ func (a *API) GetMattermostUserStatus(mattermostUserID string) (*model.Status, e
 }
 
 func (a *API) GetMattermostUserStatusesByIds(mattermostUserIDs []string) ([]*model.Status, error) {
-	return a.api.GetUserStatusesByIds(mattermostUserIDs)
+	st, err := a.api.GetUserStatusesByIds(mattermostUserIDs)
+	if err != nil {
+		return nil, err
```
mickmister: @levb If this function is implemented as just `return a.api.GetUserStatusesByIds(mattermostUserIDs)`, a nil error will be treated as a non-nil value to the caller.

You and I talked once about handling `model.AppError` and `error`, specifically that an interface pointer to a (nil pointer to a concrete type) is not a nil interface pointer, but I don't understand why this needs to be here in this case.

jfrerich:
Thanks @mickmister!  I'll get on the review soon!

server/mscalendar/availability.go
```diff
@@ -35,11 +35,25 @@ func (m *mscalendar) SyncStatusAll() (string, error) {
 		return "", err
 	}
 
-	mmIDs := userIndex.GetMattermostUserIDs()
+	index := userIndex.GetMattermostUserIDs()
```
levb: 1/5 s/index/allIDs/

server/mscalendar/availability.go
```diff
@@ -35,11 +35,25 @@ func (m *mscalendar) SyncStatusAll() (string, error) {
...
+	mmIDs := []string{}
```
levb: 1/5 s/mmIDs/filteredIDs/

server/mscalendar/oauth2.go
```diff
@@ -57,7 +74,13 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
 
 	mattermostUserID := strings.Split(state, "_")[1]
 	if mattermostUserID != authedUserID {
-		return errors.New("not authorized, user ID mismatch")
+		if mattermostUserID != app.Config.BotUserID {
```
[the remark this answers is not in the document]
mickmister: The user is checked to be authenticated [here]( https://github.com/mattermost/mattermost-plugin-mscalendar/blob/1f6b804c46a0bdad301c7c60240232e27953cb30/server/utils/oauth2connect/oauth2_connect.go#L26), before this function is called. The `authedUserID` means the user that is currently logged into MM in this request, and is not necessarily an admin. The only time where it makes sense for the current user's id to not equal the one in the oauth URL, is if we are authenticating the bot account. So if there is a mismatch, and the bot is not the user being linked, then something is wrong.

@jfrerich Does this answer your question?

server/mscalendar/oauth2.go
```diff
@@ -72,6 +95,18 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
 		return err
 	}
 
+	uid, err := app.Store.LoadMattermostUserId(me.ID)
+	if err == nil {
```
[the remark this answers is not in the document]
mickmister: When `err == nil`, that means that we asserted that the user is already connected, which is not a good thing, since we are trying to (re)connect the user. So this block is something that should reject the user's intent to reconnect to the same account. As a result, we tell them to disconnect first.

server/mscalendar/oauth2.go
```diff
@@ -72,6 +95,18 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
...
+		user, userErr := app.PluginAPI.GetMattermostUser(uid)
+		if userErr == nil {
+			app.Poster.DM(authedUserID, RemoteUserAlreadyConnected, me.Mail, user.Username)
+			return errors.Errorf(RemoteUserAlreadyConnected, me.Mail, user.Username)
+		} else {
+			// Orphaned connected account. Let it be overwritten by passing through here?
+			app.Poster.DM(authedUserID, RemoteUserAlreadyConnectedNotFound, me.Mail)
```
mickmister: @levb This is a weird situation, where the remote account has already been connected to a MM account, but the MM account failed to be retrieved from the MM server. Should we reject this connection attempt in this case? Allowing the new connection leaves the KV store in a weird state where the `mattermostUserIDKV` is overwritten, but the orphaned user still exists in `userKV` and `userIndexKV`.

I'm thinking we should reject it because the failure to retrieve the user could have been for any reason.

server/mscalendar/availability.go
```diff
@@ -35,14 +35,14 @@ func (m *mscalendar) SyncStatusAll() (string, error) {
 		return "", err
 	}
 
-	index := userIndex.GetMattermostUserIDs()
-	mmIDs := []string{}
-	for _, id := range index {
+	allIDs := userIndex.GetMattermostUserIDs()
```
levb: _only_ if you do another push anyway... a nit of a nit, but maybe inline `userIndex.GetMattermostUserIDs()` and just use `ids` for the return? :)

server/mscalendar/oauth2.go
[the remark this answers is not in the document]
jfrerich: Thanks for the clarification!

PR #35, opened 2020-02-12
WIP: [MM-22457] Delta links used in case of notification outage

mickmister:
#### Summary

Microsoft Graph's notification system has experienced outages, so we need a second solution that involves pulling. Using Graph's [Event Delta](https://docs.microsoft.com/en-us/graph/delta-query-events) queries, we can fetch only the events that have changed since last pull.

In the query response, we receive a new delta URL to use for our next request, which captures our current query and makes it so the next response is incremental off the previous query. We store these delta URLs in the local representation of a user's subscription, to use periodically throughout a given day.

These delta URLs have the entire query encoded in them, including the date range used in the query. In order to always use a practical date range, we should update the delta query once a day, moving the range start and end forward one day. This is what `InitPolling` does, whereas `Poll` runs several times a day.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22457

#### Todos

- Create job to run `InitPolling` once a day, and `Poll` every 5 minutes (or something similar)
- Make job scalable for several users, with batch requests
- Possibly delay the queuing of the simulated notification, to avoid race conditions of a real notification
- Review exposure of `NotificationProcessor` on `mscalendar` struct. It is not exposed on the interface, so other packages besides `plugin` correctly cannot access the `NotificationProcessor`

#### Additional Info

If a real notification comes in for the same event that is being processed by a delta link, there will be "no changes" to process, so the one that is processed last will be dropped, correctly. We should probably have a delay when queuing a notification from a delta link, so we avoid race conditions with a real notification.

If an event is created/updated/deleted outside of the date range used in the query, it will not be found in the delta query. When we eventually receive it using `InitPolling`, we will store it in the kvstore, but not make a DM for the user. If the event changes after it is in the kvstore, we will create a DM.

### DIFF
--- a/server/mscalendar/calendar.go
+++ b/server/mscalendar/calendar.go
@@ -9,6 +9,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/store"
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
+	"github.com/pkg/errors"
 )
 
 type Calendar interface {
@@ -18,8 +19,8 @@ type Calendar interface {
 	FindMeetingTimes(user *User, meetingParams *remote.FindMeetingTimesParameters) (*remote.MeetingTimeSuggestionResults, error)
 	GetCalendars(user *User) ([]*remote.Calendar, error)
 	ViewCalendar(user *User, from, to time.Time) ([]*remote.Event, error)
-	InitEventDelta(user *User) (events []*remote.Event, deltaURL string, err error)
-	RollingEventDelta(user *User) (events []*remote.Event, deltaURL string, err error)
+	InitPolling(user *User) (events []*remote.Event, deltaURL string, err error)
+	Poll(user *User) (events []*remote.Event, deltaURL string, err error)
 }
 
 func (m *mscalendar) ViewCalendar(user *User, from, to time.Time) ([]*remote.Event, error) {
@@ -103,25 +104,31 @@ func (m *mscalendar) GetCalendars(user *User) ([]*remote.Calendar, error) {
 	return m.client.GetCalendars(user.Remote.ID)
 }
 
-// RollingEventDelta uses the stored deltaLink in the user's subscription to fetch any events that have changed since the last fetch
+// Poll uses the stored deltaLink in the user's subscription to fetch any events that have changed since the last fetch
 // This should run frequently throughout the day to ensure we process new event updates, in case the Graph notifications are experiencing an outage
-func (m *mscalendar) RollingEventDelta(user *User) (events []*remote.Event, deltaURL string, err error) {
+func (m *mscalendar) Poll(user *User) (events []*remote.Event, deltaURL string, err error) {
 	err = m.Filter(
 		withClient,
-		withUserExpanded(user),
+		withRemoteUser(user),
 	)
 	if err != nil {
 		return nil, "", err
 	}
 
-	sub, err := m.Store.LoadUserSubscription(user.MattermostUserID)
+	sub, err := m.loadUserSubscription(user.MattermostUserID)
 	if err != nil {
 		return
 	}
 
-	events, deltaURL, err = m.client.GetEventDeltaFromURL(sub.DeltaURL)
+	if sub.PollingURL == "" {
+		return nil, "", errors.New("No polling URL stored for user " + user.Remote.Mail)
+	}
+	events, deltaURL, err = m.client.GetEventDeltaFromURL(sub.PollingURL)
+	if err != nil {
+		return nil, "", err
+	}
 
-	sub.DeltaURL = deltaURL
+	sub.PollingURL = deltaURL
 	err = m.Store.StoreUserSubscription(user.User, sub)
 	if err != nil {
 		return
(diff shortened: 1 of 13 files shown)


server/remote/msgraph/get_event_delta.go
```diff
@@ -0,0 +1,78 @@
...
+
+	u := c.rbuilder.Me().CalendarView().URL() + "/delta" + params
+	var out getEventDeltaResponse
+
+	_, err = c.CallJSON(http.MethodGet, u, nil, &out)
+	if err != nil {
+		return nil, "", err
+	}
+
+	ls := []*remote.Event{}
```
levb: s/ls/events/

server/remote/msgraph/get_event_delta.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *client) GetEventDeltaFromDateRange(remoteUserID string, start, end *remote.DateTime) (events []*remote.Event, deltaLink string, err error) {
+	q := url.Values{}
+	q.Add("StartDateTime", start.Time().Format(time.RFC3339))
+	q.Add("EndDateTime", end.Time().Format(time.RFC3339))
+	params := "?" + q.Encode()
+
+	u := c.rbuilder.Me().CalendarView().URL() + "/delta" + params
+	var out getEventDeltaResponse
+
+	_, err = c.CallJSON(http.MethodGet, u, nil, &out)
```
levb: I must be missing some details here, can you call the ...FromURL here?

server/remote/msgraph/get_event_delta.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (c *client) GetEventDeltaFromURL(deltaURL string) (events []*remote.Event, deltaLink string, err error) {
+	var out getEventDeltaResponse
+
+	_, err = c.CallJSON(http.MethodGet, deltaURL, nil, &out)
+	if err != nil {
+		return nil, "", err
+	}
+
+	ls := []*remote.Event{}
```
levb: events

server/store/subscription_store.go
```diff
@@ -21,6 +22,7 @@ type Subscription struct {
 	PluginVersion       string
 	Remote              *remote.Subscription
 	MattermostCreatorID string
+	DeltaURL            string
```
levb: PollingURL

server/store/subscription_store.go
```diff
@@ -32,6 +34,20 @@ func (s *pluginStore) LoadSubscription(subscriptionID string) (*Subscription, er
 	return &sub, nil
 }
 
+func (s *pluginStore) LoadUserSubscription(mattermostUserID string) (*Subscription, error) {
```
levb: hmm 2/5 `store` should avoid providing aggregate functions like this in its interface, this logic belongs in `mscalendar`, as a private method unless needed by other packages?

levb:
Good stuff! I recommend adding a feature flag (config?) to enable/disable the polling when you hook it all together.

server/mscalendar/calendar.go
```diff
@@ -98,3 +102,100 @@ func (m *mscalendar) GetCalendars(user *User) ([]*remote.Calendar, error) {
 
 	return m.client.GetCalendars(user.Remote.ID)
 }
+
+// RollingEventDelta uses the stored deltaLink in the user's subscription to fetch any events that have changed since the last fetch
+// This should run frequently throughout the day to ensure we process new event updates, in case the Graph notifications are experiencing an outage
+func (m *mscalendar) RollingEventDelta(user *User) (events []*remote.Event, deltaURL string, err error) {
+	err = m.Filter(
+		withClient,
+		withUserExpanded(user),
```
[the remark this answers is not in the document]
mickmister: Created `withMattermostUser` and `withRemoteUser` for clarity.

server/remote/msgraph/get_event_delta.go
mickmister: Brilliant, fits like a glove.

server/remote/msgraph/get_event_delta.go
mickmister: Habit from my course on Scheme... Good catch

server/mscalendar/calendar.go
```diff
@@ -16,6 +19,8 @@ type Calendar interface {
 	FindMeetingTimes(user *User, meetingParams *remote.FindMeetingTimesParameters) (*remote.MeetingTimeSuggestionResults, error)
 	GetCalendars(user *User) ([]*remote.Calendar, error)
 	ViewCalendar(user *User, from, to time.Time) ([]*remote.Event, error)
+	InitPolling(user *User) (events []*remote.Event, deltaURL string, err error)
```
jfrerich: Can we remove the named return types from these two new methods?  There are several occurrences in this file.

server/mscalendar/calendar.go
```diff
@@ -98,3 +103,118 @@ func (m *mscalendar) GetCalendars(user *User) ([]*remote.Calendar, error) {
...
+		withClient,
+		withRemoteUser(user),
+	)
+	if err != nil {
+		return nil, "", err
+	}
+
+	sub, err := m.loadUserSubscription(user.MattermostUserID)
+	if err != nil {
+		return
```
jfrerich: Why do we not need to return `nil, "", err` for this err case?

server/mscalendar/calendar.go
mickmister: I wanted to be clear that the `deltaURL` is what is being returned for the `string` return value, so a given dev will know what to expect when the method is called, or when reading godocs. I'm 0/5 on keeping the named return values, as there is now some inconsistency here, and the method(s) will likely not be called except in the recurring job we create. What do you think @jfrerich?

server/mscalendar/calendar.go
mickmister: It compiles because the return values are named/declared as variables in the function signature. I agree that it should have explicit values in the return statement.

server/mscalendar/calendar.go
mickmister: I'm opting for removing the `events` and `deltaURL` return values, because they are never used anyway. We can add them later if they need to be returned.

mickmister:
Closing, as this PR is just clutter at the moment. Will revisit if we experience outage issues.

PR #36, opened 2020-02-18
[MM-21140] Daily summary job

mickmister:
#### Summary

This PR allows a user to receive their daily summary of events at their configured time. The job is enabled/disabled using a system console variable. It runs every 30 minutes on the half-hour. UX around user configuration of their summary is explained in the QA steps.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21140

#### QA Steps

- Bot user must be connected with `/mscalendar connect_bot`
- Run `/mscalendar summary time 8:00AM` or whatever time in 30 minute increments that is coming up soon
- Run `/mscalendar summary enable` to enable your user's daily summary
- Enable the daily summary job in the system console
- Wait until the configured time comes up
- User should receive their daily summary of events
- If a 30 minute mark comes up and the user's time does not match, they should not receive a summary DM

#### Additional Info

Currently the time frame used to fetch the user's daily summary is `current time when summary is being sent + 24 hours`. This should be a more sophisticated query. [Ticket here](https://mattermost.atlassian.net/browse/MM-22649)

The operation takes into account the user's timezone that is set in Microsoft Calendar. We re-fetch their timezone whenever they set their configured time using `/mscalendar summary time 8:00AM` for example.


server/config/config.go
```diff
@@ -9,7 +9,8 @@ type StoredConfig struct {
 	OAuth2ClientID     string
 	OAuth2ClientSecret string
 
-	EnableStatusSync bool
+	EnableStatusSync   bool
```
levb: 👍

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+	return nil
+}
+
+func (m *mscalendar) postDailySummary(mattermostUserID string) error {
+	user := NewUser(mattermostUserID)
+	calendarData, err := m.viewTodayCalendar(user)
+	if err != nil {
+		m.Poster.DM(mattermostUserID, "Failed to run daily summary job. %s", err.Error())
```
larkox: 0/5 Maybe this message is too technical. What about "Failed to show your daily summary. Error: %s"?

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+		return err
+	}
+	if len(calendarData) == 0 {
+		m.Poster.DM(mattermostUserID, "You have no upcoming events today.")
+	} else {
+		m.Poster.DM(mattermostUserID, utils.JSONBlock(calendarData))
```
larkox: I understand we might be doing this as a placeholder for the future formatting of the calendar data. Should not we already use a function like "formatCalendar" or "printCalendar" and then there call the utils.JSONBlock? That way when we finally implement that formatting, we do not have to track all these places where the calendarData is printed.

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return false, err
+	}
+	t, err := time.ParseInLocation(time.Kitchen, dsum.PostTime, loc)
+	if err != nil {
+		return false, err
+	}
+
+	now := timeNowFunc().In(loc)
+	if now.Weekday() == time.Saturday || now.Weekday() == time.Sunday {
```
larkox: I see changing this a future feature (not everyone's week is Monday to Friday). Maybe a good help-wanted ticket?

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	err = m.Store.SaveDailySummaryIndex(dsumIndex)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func (m *mscalendar) DailySummaryAll() error {
```
larkox: Naming here feels weird. I would go for a "PostAllDailySummary" or just "PostDailySummary".

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+import (
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/store"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/tz"
+	"github.com/pkg/errors"
+)
+
+const dailySummaryTimeWindow = time.Minute * 2
```
larkox: If the DailySummaryJob deals with calling the DailySummaryAll every 30 minutes, couldn't we be more generous here? Maybe up to 20 minutes?

server/store/daily_summary_store.go
```diff
@@ -0,0 +1,39 @@
...
+
+type DailySummaryStore interface {
+	LoadDailySummaryIndex() (DailySummaryIndex, error)
+	SaveDailySummaryIndex(dsumIndex DailySummaryIndex) error
+}
+
+type DailySummarySettings struct {
+	MattermostUserID string `json:"mm_id"`
+	Enable           bool   `json:"enable"`
+	PostTime         string `json:"post_time"` // Kitchen format, i.e. 8:30AM
```
larkox: I would add a LastPostTime field to avoid repetition over different clusters on HA. It is not optimal solution, but may help.

server/mscalendar/daily_summary.go
larkox: This is specially important for the (rare) timezones where they have an offset of 15 minutes, instead of a full hour.

server/mscalendar/daily_summary_test.go
```diff
@@ -0,0 +1,73 @@
...
+	dsum.Timezone = "Mountain Standard Time"
+	shouldRun, err = shouldPostDailySummary(dsum)
+	require.NotNil(t, err)
+	require.False(t, shouldRun)
+
+	dsum.PostTime = "7:00AM"
+	dsum.Timezone = "Moon Time" // Invalid timezone
+	shouldRun, err = shouldPostDailySummary(dsum)
+	require.NotNil(t, err)
+	require.False(t, shouldRun)
```
larkox: I would add a test with "Nepal Standard Time" (UTC+5:45). It is a weird case, but it is worth keeping in mind I think.

server/store/store.go
```diff
@@ -19,6 +19,7 @@ const (
 	OAuth2KeyPrefix           = "oauth2_"
 	SubscriptionKeyPrefix     = "sub_"
 	EventKeyPrefix            = "ev_"
+	DelaySummaryKeyPrefix     = "dsum_"
```
larkox: typo? DailySummary?

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+	if t.Minute() != 0 && t.Minute() != 30 {
+		return nil, errors.New("Time must be a multiple of 30 minutes.")
+	}
+
+	timezone, err := m.GetTimezone(user)
+	if err != nil {
+		return nil, err
+	}
+
+	dsumIndex, err := m.Store.LoadDailySummaryIndex()
```
larkox: During these actions, shouldn't we use KVCompareAndSet to avoid problems with HA?

server/mscalendar/daily_summary_job.go
```diff
@@ -0,0 +1,98 @@
...
+	j.Logger.Debugf("Daily summary job finished")
+}
+
+func (j *DailySummaryJob) timerUntilFirstRun() *time.Timer {
+	now := time.Now()
+	var target time.Time
+	if now.Minute() < 30 {
+		target = time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, time.Local).Add(30 * time.Minute)
+	} else {
+		target = time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, time.Local).Add(60 * time.Minute)
```
jfrerich: make this statement the predefined value and remove the `else`?

plugin.json
```diff
@@ -86,6 +86,13 @@
                 "type": "bool",
                 "help_text": "When enabled, a Mattermost user's status will automatically update based on their Microsoft Calendar availability. This runs every 5 minutes.",
                 "default": false
+            },
+            {
+                "key": "EnableDailySummary",
+                "display_name": "Enable Daily Summary Job",
```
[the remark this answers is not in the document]
mickmister: This is in the admin console, which is why I worded it as `Job`. I think it's important to know that it is a opt-in job running on their hardware. 0/5 on keeping it as `Job`, or only referencing it as `Job` in the config value's `help_text`, which is next to the name of the config value in the admin console.

cc @aaronrothschild

server/command/daily_summary.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func (c *Command) dailySummary(parameters ...string) (string, error) {
+	if len(parameters) == 0 {
+		dsum, err := c.MSCalendar.GetDailySummarySettingsForUser(c.user())
+		if err != nil {
+			return fmt.Sprintf("Error: %s", err.Error()), nil
+		}
+
+		return utils.JSONBlock(dsum), nil
```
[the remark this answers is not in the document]
mickmister: I agree that it should be better than JSON. I'll format it better and move the formatting code to its own function in this PR if you would like.

server/mscalendar/daily_summary.go
mickmister: I'm weary of using `Post` in the name, because most of the time the job runs, there will be no posts. It makes me think we are force-posting to all users, regardless if it is at their saved time. @levb Do you have any thoughts on this?

server/mscalendar/daily_summary.go
mickmister: This time computation algorithm will likely change with the shared timing logic in the cluster job PR. 

Do you think we should enable the users to set their times at 15 minute increments, or keep them at 30 minute increments?

server/mscalendar/daily_summary.go
mickmister: Ticket here https://github.com/mattermost/mattermost-plugin-mscalendar/issues/47

server/mscalendar/calendar.go
```diff
@@ -29,6 +29,24 @@ func (m *mscalendar) ViewCalendar(user *User, from, to time.Time) ([]*remote.Eve
...
+		return nil, err
+	}
+
+	err = m.ExpandRemoteUser(user)
+	if err != nil {
+		return nil, err
+	}
+
+	from := time.Now()
+	to := time.Now().Add(time.Hour * 24)
```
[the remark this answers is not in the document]
mickmister: I think I will push an update that implements the final scenario you suggested.

server/mscalendar/daily_summary.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	return nil, errors.New("No daily summary settings found")
+}
+
+func (m *mscalendar) SetDailySummaryPostTime(user *User, timeStr string) (*store.DailySummarySettings, error) {
+	t, err := time.Parse(time.Kitchen, timeStr)
+	if err != nil {
+		return nil, errors.New("Invalid time value")
```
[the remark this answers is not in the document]
mickmister: The function that calls this function is in the `command` package, which is more user-facing than the `mscalendar` package. I've added the tip of a valid values in the `command` package function.

mickmister:
@larkox I've addressed your feedback, besides the timezones that have 15 minute differences, and the `Monday to Friday` complications with some people having weekdays other than that.

I've made a ticket to handle the concurrent `KVCompareAndSet` issues [here](https://github.com/mattermost/mattermost-plugin-mscalendar/issues/47).

server/mscalendar/daily_summary.go
levb: Not sure of the full context, but maybe `Generate...`?

server/mscalendar/daily_summary.go
larkox: If we put them at 15 minutes, we may solve the problem with the timezones, but I have no strong opinion on this.

server/mscalendar/daily_summary_test.go
mickmister: Test case added!

server/mscalendar/daily_summary.go
mickmister: I decided to make the job interval 15 minutes, as well as allow the user to set a time divisible by 15 minutes.

server/mscalendar/daily_summary.go
mickmister: Ticket here https://github.com/mattermost/mattermost-plugin-mscalendar/issues/56

server/mscalendar/calendar.go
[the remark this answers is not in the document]
mickmister: @larkox Pushed an update with your final suggestion here

mickmister:
@larkox All of your feedback should be addressed with the latest commits

server/mscalendar/daily_summary.go
larkox: (0/5) I still would be a bit worried if the job gets desynchronized and no new summaries being sent (the 2 minute time window seem a bit narrow). Nevertheless, if there is a problem with desynchronization, it does not matter how big is the time window, eventually there will be a problem either way.

server/mscalendar/calendar.go
```diff
@@ -32,7 +29,7 @@ func (m *mscalendar) ViewCalendar(user *User, from, to time.Time) ([]*remote.Eve
 	return m.client.GetDefaultCalendarView(user.Remote.ID, from, to)
 }
 
-func (m *mscalendar) viewTodayCalendar(user *User) ([]*remote.Event, error) {
+func (m *mscalendar) viewTodayCalendar(user *User, timezone string) ([]*remote.Event, error) {
```
larkox: View sounds like "this will be printed on the screen". Wouldn't it be better "getTodayCalendar" or something similar?

mickmister: Added the change :+1:

DHaussermann:
Testing done. No functional issues found.
- Events are delivered on time interval where they are set
- Events are not delivered on time intervals where they are not set
- No issues updating delivery time    
- Adjusting timezone is ms calendar is reflected on next set 
- Job can be set to 15 minutes interval (not just 30 or on the hour)
- View is functional   
- Settings is functional 
- Disable the job on system level but have enable for user --> No summary 
- Enable the Job on system level but have it disabled for the user --> No summary 

@mickmister  some issues to review. (1. and 2. may be out of scope of this PR as they are about `view`  
1. There was one case where where an appointment was not returned in the data. At first I thought there was logic to omit events that have already passed but then I realized there is not.  I'm confused why this one appointment does not come back with view. I see nothing unique about it on the MS side.
Do you do any kind of filtering on this data?      
MissingAppointement

2. /`mscalendar summary view` does not post in any channel other channel other than MS Calendar Bot DM. When I run the command from any other channel the reply is always in the Bot DM channel. This feels like a bug to me. It's confusing and inconsistent with the other commands that reply in the channel where the slash command is posted. 

3. Minor changes to text from MS Calendar posts 
- `Error: User not connected: not found` when logged in user is mapped
- Should `/mscalendar summary settings` show full help? when I see the full list of commands shown, the first impression is that the command did not work. I feel like the help should not be shown. The same way the other commands don't show the help.

mickmister:
@DHaussermann Thanks for the review.

I'm currently trying to figure out `1.` I'm not sure what could cause the event not to show up.

Regarding `3. Error: User not connected: not found`, was this when the bot wasn't connected?

I've addressed the rest of your feedback in the latest commit.

mickmister:
@DHaussermann I've addressed your feedback regarding to the two points in my previous comment.

The time range for the events in the daily summary is now
`midnight of current day (beginning of current day)` - `midnight of next day (end of current day)`. This behavior was agreed on offline with @DHaussermann and @aaronrothschild

levb:
@mickmister This doesn't appear to be using the new cluster API, like the status sync job. Why? Or am missing something?

mickmister:
@levb I'm following up with the a PR regarding batch requests, where I'm also taking care of the cluster job. I was trying not to add more scope to this, even though in retrospect changing to use the cluster job is not a big change and seems a better fit for this PR.

@DHaussermann This PR is blocking another PR and should be ready for re-review

PR #38, opened 2020-02-24
Fix findmeetings typo

MatthewDorner:
#### Summary
There is a typo in the autocomplete description for the `/mscalendar` command, and since the command is case-sensitive it will not work if typed as shown. It should be `findmeetings`. I didn't find any other instances of this typo in the repository.

### DIFF
--- a/server/command/command.go
+++ b/server/command/command.go
@@ -45,7 +45,7 @@ func Register(registerFunc RegisterFunc) {
 		DisplayName:      "Microsoft Calendar",
 		Description:      "Interact with your outlook calendar.",
 		AutoComplete:     true,
-		AutoCompleteDesc: "info, connect, disconnect, connect_bot, disconnect_bot, subscribe, showcals, viewcal, createcal, deletecal, createevent, findmeetings, availability",
+		AutoCompleteDesc: "help, info, connect, disconnect, connect_bot, disconnect_bot, subscribe, showcals, viewcal, createcal, deletecal, createevent, findmeetings, availability",
 		AutoCompleteHint: "(subcommand)",
 	})
 }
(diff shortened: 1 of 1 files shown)


hanzei:
@mickmister I noticed that some of the commands found in https://github.com/mattermost/mattermost-plugin-mscalendar/blob/master/server/command/command.go#L61-L88 are not listed in the hint. Should they be added?

mickmister:
@hanzei Yes, it seems `disconnect`, `connect_bot`, and `disconnect_bot` are missing. `testcreateevent` should be listed as `createevent`, as well.

@MatthewDorner Do you mind including these in your changes here?

MatthewDorner:
There is no help command either, or descriptions of the parameters the commands accept such as `subscribe list | show | renew | delete `.

mickmister:
@MatthewDorner How about something like this:

`subscribe (list | show | renew | delete), ...`

These would also need to show up in the output of `/mscalendar help`.

We can break this out into a separate ticket since it is growing in scope. What do you think @MatthewDorner?

MatthewDorner:
We can make it a separate issue. I will just add the missing commands for this PR.

MatthewDorner:
There is a `getHelp()` placeholder function in `command.go`: https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/server/command/command.go#L29
This led me to think there was no help text, but actually it's just implemented elsewhere in `command/help.go` (even though it needs to be updated as noted in this thread).

Anyway if `getHelp()` won't be used, maybe it should be removed from the code.

mickmister:
@MatthewDorner I think it makes sense to have `getHelp()` implement the help text, and have `c.help` call `getHelp()` in order to use the help text. 1/5 That's fine to be done in this PR.

mickmister:
Tested, this works

Issue #13, opened 2019-12-11
Generate cilentState (a secret) when a new subscription is created


levb:
Microsoft Graph webhooks use the field `ClientState` for authenticating webhook notification messages on subscriptions. Presently, the POC code verifies the secret if it were set, but does not set it.

A simple submission will create (and store) the secret in in https://github.com/mattermost/mattermost-plugin-msoffice/blob/80abaaaa99f419dbbfe6555948f7ccca0bc5ecd2/server/api/subscription.go#L25

A more comprehensive PR will include tests for webhook message validation using `remote` API mocks as needed.

Original Jira ticket: https://mattermost.atlassian.net/browse/MM-21141

MatthewDorner:
I would like to try this one.

PR #40, opened 2020-03-01
Add ClientState and tests

MatthewDorner:
#### Summary
- Generate random `ClientState` when creating a new subscription
- Tests to verify that `ClientState` is checked for incoming webhooks
- Fixed bug in `mscalendar/notification.go` involving conditional call to `MakeClient`
- Fixed logging typos

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/13


MatthewDorner:
I had to copy the definition of the `webhook` struct into my test since it's defined in the `msgraph` package and isn't exported, and I need it for my test `remote.Notification` object. What's the correct solution to this? Just change it to be exportable?

mickmister:
> I had to copy the definition of the webhook struct type into my test since it's defined in the msgraph package (in remote/msgraph/handle_webhook.go) and isn't exported, and I need it for my test remote.Notification object. What's the correct solution to this? Just change it to be exported?

@MatthewDorner We can create a mock of `remote.Notification` to be used in the tests. We will need to add a line [here](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/Makefile#L100) to generate the mock, then run `make mock` in your terminal. Then we can use the mock in tests.

[Example init](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/server/mscalendar/oauth2_test.go#L484)

[Example usage](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/server/mscalendar/oauth2_test.go#L320)

MatthewDorner:
Never mind, I guess I didn't even need that field for my test.

MatthewDorner:
Should I also be using `MockLogger` in the test? It's outputting log messages into the test output, so I'm guessing I should mock it. I just used `&bot.TestLogger{TB: t}` since I saw it in other tests.

mickmister:
@MatthewDorner I would say that you don't need to mock it unless you want to check the logger's output. This can cause brittle tests that need to be unnecessarily updated if there are more logging additions later.

I think the difference between `TestLogger` and `NilLogger` is whether or not you want the logs to show up in the test run's output. 0/5 `NilLogger` sounds like a fine choice.

server/mscalendar/notification_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return processor
+}
+
+func TestProcessNotification(t *testing.T) {
+	token := &oauth2.Token{
+		AccessToken: "creator_oauth_token",
+	}
+
+	eRemote := &remote.Event{
```
mickmister: Do you think each of these test structs can be moved to functions such as `newTestEvent`?

server/mscalendar/notification_test.go
```diff
@@ -0,0 +1,190 @@
...
+				Dependencies: &Dependencies{
+					Store:     mockStore,
+					Logger:    logger,
+					Poster:    mockPoster,
+					Remote:    mockRemote,
+					PluginAPI: mockPluginAPI,
+				},
+			}
+
+			nRemote.ClientState = tc.incomingClientState
```
mickmister: 1/5 I suggest that instead of creating one `notification` in the test init and editing it here, to create them in the per-test `tc` structs above.

server/mscalendar/notification_test.go
```diff
@@ -0,0 +1,190 @@
...
+			nRemote.RecommendRenew = tc.recommendRenew
+
+			nRemoteWithEvent := *nRemote
+			nRemoteWithEvent.Event = eRemote
+
+			mockStore.EXPECT().LoadSubscription("remote_subscription_id").Return(sStore, nil).Times(1)
+			mockStore.EXPECT().LoadUser("creator_mm_id").Return(uStore, nil).Times(1)
```
mickmister: The name `uStore` makes me think the variable is an instance of `UserStore`. 1/5 I think the variable name `user` is sufficient. Other similar variable names should probably be changed as well.

server/mscalendar/notification_test.go
```diff
@@ -0,0 +1,190 @@
...
+					mockStore.EXPECT().StoreUserSubscription(uStore, &store.Subscription{
+						Remote:              &remote.Subscription{},
+						MattermostCreatorID: "creator_mm_id",
+						PluginVersion:       "x.x.x",
+					}).Return(nil).Times(1)
+				}
+
+				mockClient.EXPECT().GetNotificationData(nRemote).Return(&nRemoteWithEvent, nil).Times(1)
+
+				if tc.hasPriorEvent {
```
mickmister: Maybe make a field `tc.priorEvent store.Event`, and create a new one for each `tc`

server/mscalendar/notification_test.go
```diff
@@ -0,0 +1,190 @@
...
+					}, nil).Times(1)
+				} else {
+					mockStore.EXPECT().LoadUserEvent("creator_mm_id", "remote_event_id").Return(nil, store.ErrNotFound).Times(1)
+				}
+
+				mockPoster.EXPECT().DMWithAttachments("creator_mm_id", gomock.Any()).Return(nil).Times(1)
+				mockStore.EXPECT().StoreUserEvent("creator_mm_id", gomock.Any()).Return(nil).Times(1)
+			}
+
+			p := NewTestNotificationProcessor(env)
```
mickmister: 0/5 to make this a private function

server/remote/msgraph/subscription.go
```diff
@@ -14,11 +16,16 @@ import (
 const subscribeTTL = 10 * time.Minute
 
 func (c *client) CreateMySubscription(notificationURL string) (*remote.Subscription, error) {
+	b := make([]byte, 96)
+	rand.Read(b)
+	clientState := base64.URLEncoding.EncodeToString(b)
```
mickmister: 1/5 put this logic in its own function `newRandomString`

MatthewDorner: Should it be in the same file?

mickmister: Sounds good to me :+1:

server/remote/msgraph/subscription.go
```diff
@@ -13,12 +15,19 @@ import (
 
 const subscribeTTL = 10 * time.Minute
 
+func newRandomString() string {
```
levb: Just out of curiosity, why not a new guid?

MatthewDorner: A GUID wouldn't be a full 128 characters, and I thought there might be some advantage in security or randomness by not using a GUID. On the other hand, a GUID should technically work.

levb: They are 128 bits, often encoded as a 32-character hex value.

MatthewDorner: The `ClientState` can be up to 128 characters so I thought I'd use the full length, although it will accept a shorter string.

MatthewDorner: (base64 will encode 96 bytes into a 128-character string)

MatthewDorner: And for the reasons listed here: https://devblogs.microsoft.com/oldnewthing/20150701-00/?p=45241

MatthewDorner:
Anything else needed for this PR? I believe all requested changes should be addressed.

mickmister:
@DHaussermann Testing here will be a smoke test for setting up a subscription and verifying that we can successfully process a notification.

mickmister:
@DHaussermann I think we're good to merge this and test on master

PR #42, opened 2020-03-03
Delete subscription upon disconnecting user

### LINKED ISSUE #34: Delete subscriptions upon user disconnect
#### Summary

Currently, the disconnect functionality only removes the user from the KV store, and does not do anything on Microsoft's side. If the user has an active subscription, we should delete it in the KV store, as well as on Microsoft's side so we don't receive unnecessary webhook events.

Info on deleting a subscription using the Microsoft Graph API: https://docs.microsoft.com/en-us/graph/api/subscription-delete?view=graph-rest-1.0&tabs=http

MatthewDorner:
#### Summary
Add code in `DisconnectUser` to delete the user's event subscription from the KV store and also from the MS Graph API. Let me know if I should change the error handling or messages. I tried to imitate the error handling from `mscalendar/subscription.go` where subscriptions are deleted, and I decided not to return error if the MS Graph API deletion fails since the subscription may have expired.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/34


mickmister (changes requested):
@MatthewDorner Great job on the implementation! Just one suggestion on logging an error.

levb (dismissed):

mickmister (dismissed):
LGTM, thanks @MatthewDorner!

mickmister:
@DHaussermann I'll defer to you whether you want to test this before master

hanzei:
@DHaussermann Gentle reminder to review this PR :wink:

hanzei:
@MatthewDorner Heads up that there is a merge conflict to resolve

MatthewDorner:
Oh, I didn't notice the merge conflict. Will fix.

MatthewDorner:
I'm guessing tests are failing until https://github.com/mattermost/mattermost-plugin-mscalendar/pull/63 gets merged in.

hanzei:
That is correct @MatthewDorner. Sorry for the inconvenience.

DHaussermann:
@MatthewDorner I'm not seeing this work as expected. I wonder if the way I'm trying to validate this is correct?
Steps:
- Query `PluginKeyValueStore` and see how many MS Calendar records have a `PKey` value of `sub_...`
- In a browser session, connect test user with the plugin
- Note that one additional sub record is visible in the ``PluginKeyValueStore` table (as expected)
- Return to browser session and disconnect test user
- Return to DB browser and query table again
**Expected**: Row count is reduced by 1
**Observed**: Row count is unchanged. Nothing written to the log by the plugin.

Is this maybe dependant on another change in master? Thoughts?

MatthewDorner:
KV Store

What I'm seeing is that when you run `/mscalendar connect` it creates row 8 in this screenshot. Then when you run `/mscalendar subscribe`, it creates row 7, and when you run `/mscalendar disconnect` with the changes in this PR, it deletes row 7. So it creates two records but deletes one.

I'm not sure what the value in row 8 is supposed to be.

hanzei:
@MatthewDorner Build is green now

MatthewDorner:
I think the record that's not getting deleted is the OAuth state being saved here: https://github.com/mattermost/mattermost-plugin-mscalendar/blob/fdde4b5ed36c87380a3cc03923150efad97747df/server/store/oauth2_store.go#L27

Maybe it shouldn't be saved with the `sub_` prefix in the first place, since it's not a subscription? And I'm not sure how to delete it, although it should probably be deleted upon `/mscalendar disconnect` since it gets created upon `/mscalendar connect`. Should this be a separate issue?

DHaussermann:
Thanks @MatthewDorner you're correct. The record I was seeing was caused by the oAuth connection. 
I was not seeing the additional record as I was not re-subscribing.

DHaussermann (approved):
Tested and passed.
Confirmed that subscription record is in fact removed from KV store when user is disconnected. 
LGTM! Thanks @MatthewDorner  for the PR!

Issue #37, opened 2020-02-23
Nil pointer dereference when user from Graph API doesn't have Mail field


MatthewDorner:
https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/server/remote/msgraph/get_me.go#L17
Panic occurs when the user retrieved from Graph API doesn't have a `Mail` field. I got this error when setting up the plugin for development, since I have Azure AD but I don't have Office 365 attached to it and my user has no e-mail.

```
DEBU[2020-02-20T16:38:43.831909-06:00] Received HTTP request                         caller="web/handlers.go:79" method=POST request_id=gxaz1dddcfyaxribxfsgjrg8mc url=/api/v4/channels/members/me/view
DEBU[2020-02-20T16:38:45.9117403-06:00] Received HTTP request                         caller="web/handlers.go:79" method=POST request_id=5rft8t3mftn6xqk9xhotyshiya url=/api/v4/channels/members/me/view
DEBU[2020-02-20T16:38:47.1997569-06:00] panic: runtime error: invalid memory address or nil pointer dereference  caller="plugin/hclog_adapter.go:53" plugin_id=com.mattermost.mscalendar
DEBU[2020-02-20T16:38:47.1998222-06:00] [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd09480]  caller="plugin/hclog_adapter.go:53" plugin_id=com.mattermost.mscalendar
DEBU[2020-02-20T16:38:47.199837-06:00]                                               caller="plugin/hclog_adapter.go:53" plugin_id=com.mattermost.mscalendar
DEBU[2020-02-20T16:38:47.1998491-06:00] goroutine 14 [running]:                       caller="plugin/hclog_adapter.go:53" plugin_id=com.mattermost.mscalendar
DEBU[2020-02-20T16:38:47.199864-06:00] github.com/mattermost/mattermost-plugin-mscalendar/server/remote/msgraph.(*client).GetMe(0xc0003ab980, 0x11d5ea0, 0xc000024130, 0xc0002d8ae0)  caller="plugin/hclog_adapter.go:53" plugin_id=com.mattermost.mscalendar
[...]

PR #52, opened 2020-03-11
Handle case where mail is nil

larkox:
#### Summary
Handle the situation where msgraph returns a nil mail, returning an error saying that there is no mail associated to the user. 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/37


mickmister:
@larkox If we're checking for all of these fields, we'd also want to update them if we have more required fields. I'm fairly certain that `Mail` is the only one we need to worry about, because it is contingent on having Outlook setup, and the others seem to be there by default. Either way, it's good to have informative errors so this is good :+1:

server/remote/msgraph/get_me.go
```diff
@@ -3,13 +3,31 @@
...
+)
 
 func (c *client) GetMe() (*remote.User, error) {
 	graphUser, err := c.rbuilder.Me().Request().Get(c.ctx)
 	if err != nil {
 		return nil, err
 	}
+
+	if graphUser.ID == nil {
+		return nil, fmt.Errorf("User has no ID")
```
mickmister: Do you mind using `errors.New` since we are not injecting any values?

```suggestion
		return nil, errors.New("User has no ID")
```

server/remote/msgraph/get_me.go
```diff
@@ -3,13 +3,31 @@
...
+		return nil, errors.New("User has no ID")
+	}
+	if graphUser.DisplayName == nil {
+		return nil, errors.New("User has no Display Name")
+	}
+	if graphUser.UserPrincipalName == nil {
+		return nil, errors.New("User has no Principal Name")
+	}
+	if graphUser.Mail == nil {
+		return nil, errors.New("User has no mail")
```
mickmister: On second thought, I think we should say something like this:
```suggestion
		return nil, errors.New("User has no email address. Please connect the account to Outlook.")
```

Also, does this error show to the user after returning?

larkox: GetMe, at the moment, is only used on the `CompleteOAuth2` function. This function returns an error that eventually will be processed by `WriteJSONError`, which will write the error on the web response, showing it to the user.
That being said, maybe the message should be more clear, since the may already think they are "connecting the account to Outlook". This error, if I remember correctly, is because the microsoft account is not connected to an outlook account, right?
What do you think about "User has no email address. Make sure the Microsoft account is associated to an Outlook product."

mickmister: > "User has no email address. Make sure the Microsoft account is associated to an Outlook product."

Sounds good to me!

Issue #57, opened 2020-03-17
Send notification for an upcoming event


larkox:
Receive a notification 10 minutes before an event takes place.

PR #58, opened 2020-03-17
[GH57] Receive notifications of upcoming events

larkox:
#### Summary
Receive notifications from upcoming events 10 minutes before the event.

Right now we are using the same job as StatusSync to avoid code repetition (the process is almost the same). Nevertheless, it can be decoupled.

Since I do not know the approach we want to take, I delayed the changes in the naming until we make a decision on whether to separate this from the status sync or keep it together.

IMPORTANT: The only way to show the subject and location of the meeting right now requires the user to explicitly give the bot account access to the user's calendar within Outlook. If not, no subject will be present.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/57


server/mscalendar/availability.go
```diff
@@ -124,6 +128,49 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+	message := "You have an upcoming event:"
+	start, err := time.Parse("2006-01-02T15:04:05MST", s.Start.DateTime+s.Start.Timezone)
+	if err != nil {
+		m.Logger.Errorf("problem parsing start date: %s", err.Error())
+		return
+	}
+
+	end, err := time.Parse("2006-01-02T15:04:05MST", s.End.DateTime+s.End.Timezone)
+
+	message = fmt.Sprintf("\n%s-%s", start.Format("15:04MST"), end.Format("15:04MST"))
```
mickmister: Unfortunate that there is not a const for `15:04MST`. Which got me thinking, instead of showing the timezone of the event in the output, I suggest we show the times in the user's timezone. The function signature here could be `renderScheduleItem(s remote.ScheduleItem, timezone string)`

Grab the user's current timezone from a call to `GetUserTimezone`, then pass it into `renderScheduleItem`. Then use it like:

```go
start := s.Start.In(timezone).Time()
end := s.End.In(timezone).Time()
message = fmt.Sprintf("\n%s-%s", start.Format(time.Kitchen), end.Format(time.Kitchen))
```

server/mscalendar/availability.go
```diff
@@ -124,6 +128,49 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+	if s.Subject == "" {
+		message += fmt.Sprintf("Subject: Not available. Check your privacy settings so we can show you the subject.")
```
mickmister: I like this method of error handling :+1: Is it possible that the event simply doesn't have a subject? I'm thinking Outlook wouldn't allow that.

mickmister:
@aaronrothschild If there is no subject on the event, then we assume that the user's settings in Outlook are set to not share the subject. @larkox has found that having the subject be private is the default config with an Outlook account. See his note in the PR description.

@aaronrothschild Two topics I want to discuss:

1. Wondering what you think about the warning about the subject showing in the event output itself. It will show up as `Subject: Not available. Check your privacy settings so we can show you the subject.` I like this approach because it says "hey the subject could be here but it's not right now."

2. How do we educate the users that they may need to change their privacy settings to share their schedule information with the bot user (so they can see the event subject in the output)? Sharing the calendar doesn't look like something we can do through Microsoft's API, so we I suppose we can have screenshots in the readme on how to share it with the bot.

@larkox Can you post a screenshot of what it looks like when they get the privacy setting warning in the event reminder?

server/mscalendar/availability.go
```diff
@@ -124,6 +128,49 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
 }
 
+func (m *mscalendar) notifyUpcomingEvent(mattermostUserID string, items []remote.ScheduleItem) {
+	for _, scheduleItem := range items {
+		loc, err := time.LoadLocation(scheduleItem.Start.Timezone)
+		if err != nil {
+			m.Logger.Errorf("problem loading location: %s", err.Error())
+			continue
+		}
+		now := time.Now().In(loc)
```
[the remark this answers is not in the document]
larkox: In order to do this right, we would have to make the assumption that start.Time() and time.Now() will be in the same location. It is safe to make that assumption?

server/mscalendar/availability.go
larkox: Apparently, you can have an event without subject. I will change the message to something more general like:

"Subject: Not available. The event may not have a subject or there might be a problem with the permissions. If you believe there is a permissions problem, check your privacy settings so we can show you the subject."

larkox: @aaronrothschild thoughts on my previous comment?

server/mscalendar/availability.go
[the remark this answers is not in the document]
larkox: I realized that with comparisons, the location makes no difference. It only makes difference for formatting the result. Thanks!

server/mscalendar/availability.go
```diff
@@ -124,6 +128,49 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+			m.Logger.Errorf("problem parsing date: %s", err.Error())
+			continue
+		}
+		if now.Add(upcomingEventNotificationTime).After(start.Add(-upcomingEventNotificationWindow)) && now.Add(upcomingEventNotificationTime).Before(start.Add(upcomingEventNotificationWindow)) {
+			m.renderScheduleItem(mattermostUserID, scheduleItem)
+		}
+	}
+}
+
+func (m *mscalendar) renderScheduleItem(mattermostUserId string, s remote.ScheduleItem) {
```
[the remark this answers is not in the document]
larkox: Having all the views in one place seems a great idea.

larkox:
@mickmister With the new proposed message it will look like this:
Screenshot from 2020-03-30 16-24-31

aaronrothschild:
> 1. Wondering what you think about the warning about the subject showing in the event output itself. It will show up as `Subject: Not available. Check your privacy settings so we can show you the subject.` I like this approach because it says "hey the subject could be here but it's not right now."

@mickmister How about: `Check your privacy settings in Outlook so we can display the subject of the meeting.`. Where exactly are the settings?

> 2. How do we educate the users that they may need to change their privacy settings to share their schedule information with the bot user (so they can see the event subject in the output)? Sharing the calendar doesn't look like something we can do through Microsoft's API, so we I suppose we can have screenshots in the readme on how to share it with the bot.

Screenshots in Readme, but also - the initial connection wizard should tell the user what to do to change their settings proactively.  

> 
> @larkox Can you post a screenshot of what it looks like when they get the privacy setting warning in the event reminder?

larkox:
@aaronrothschild I will add a message on the login Workflow about this. What about this text?
> In order to show all the information of your schedule, you need to share your calendar with the bot account. To do so, in the calendar view hit the `Share` button and share it with `bot@example.com`.

Regarding the text in case the subject is missing, cannot be so direct, because the event may actually have no subject. Here is once more the text I propose, in case you like it (in my previous comment you can see an image on how it would look):
> Subject: Not available. The event may not have a subject or there might be a problem with the permissions. If you believe there is a permissions problem, check your privacy settings so we can show you the subject.

server/mscalendar/user.go
```diff
@@ -95,6 +95,23 @@ func (m *mscalendar) GetTimezone(user *User) (string, error) {
 	return settings.TimeZone, nil
 }
 
+func (m *mscalendar) GetTimezoneByID(mattermostUserID string) (string, error) {
+	c, err := m.MakeSuperuserClient()
```
mickmister, line 99:
```suggestion
func (m *mscalendar) GetTimezoneByUserID(mattermostUserID string) (string, error) {
	return m.GetTimezone(NewUser(mattermostUserID))
```

server/mscalendar/availability.go
```diff
@@ -124,6 +130,37 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+		upcomingTime := time.Now().Add(upcomingEventNotificationTime)
+		start := scheduleItem.Start.Time()
+		diff := start.Sub(upcomingTime)
+
+		if (diff < upcomingEventNotificationWindow) && (diff > -upcomingEventNotificationWindow) {
+			var err error
+			if timezone == "" {
+				timezone, err = m.GetTimezoneByID(mattermostUserID)
+				if err != nil {
+					m.Logger.Errorf("error notifying upcoming event, err=%s", err.Error())
```
mickmister, line 145:
```suggestion
					m.Logger.Errorf("notifyUpcomingEvent error getting timezone, err=%s", err.Error())
```

server/mscalendar/availability.go
```diff
@@ -124,6 +130,37 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+					return
+				}
+			}
+
+			message, err := views.RenderScheduleItem(scheduleItem, timezone)
+			if err != nil {
+				m.Logger.Errorf("error notifying upcoming event, err=", err.Error())
```
mickmister, line 152:
```suggestion
				m.Logger.Errorf("notifyUpcomingEvent error getting rendering schedule item, err=", err.Error())
```

server/mscalendar/availability.go
```diff
@@ -124,6 +130,37 @@ func (m *mscalendar) setUserStatuses(filteredUsers store.UserIndex, schedules []
...
+				continue
+			}
+			err = m.Poster.DM(mattermostUserID, message)
+			if err != nil {
+				m.Logger.Errorf("error notifying upcoming event, err=", err.Error())
```
mickmister, line 157:
```suggestion
				m.Logger.Errorf("notifyUpcomingEvent error creating DM, err=", err.Error())
```

larkox:
Since we decided to drop the bot and all events will receive the correct subject, I will drop the task for adding the "share the calendar with the bot" message to the login flow, and will simplify how the message is shown when there is no subject.

server/mscalendar/availability.go
```diff
@@ -106,6 +112,37 @@ func (m *mscalendar) setUserStatuses(users store.UserIndex, schedules []*remote.
 	return utils.JSONBlock(schedules), nil
 }
 
+func (m *mscalendar) notifyUpcomingEvent(mattermostUserID string, items []remote.ScheduleItem) {
+	var timezone string
+	for _, scheduleItem := range items {
```
mickmister, line 117: I was going to suggest adjusting this function to take one `remote.ScheduleItem`, and have the caller loop through the different items (and pass in the timezone). But the strategy here with lazy loading the timezone seems worth it, considering how often this job runs.

server/mscalendar/availability.go
```diff
@@ -106,6 +112,37 @@ func (m *mscalendar) setUserStatuses(users store.UserIndex, schedules []*remote.
...
+		upcomingTime := time.Now().Add(upcomingEventNotificationTime)
+		start := scheduleItem.Start.Time()
+		diff := start.Sub(upcomingTime)
+
+		if (diff < upcomingEventNotificationWindow) && (diff > -upcomingEventNotificationWindow) {
+			var err error
```
mickmister, line 123: Any reason why we need to declare this here?

server/mscalendar/views/calendar.go
```diff
@@ -64,3 +64,22 @@ func groupEventsByDate(events []*remote.Event) [][]*remote.Event {
 	}
 	return result
 }
+
+func RenderScheduleItem(s remote.ScheduleItem, timezone string) (string, error) {
+	message := "You have an upcoming event:"
+	start := s.Start.In(timezone).Time()
+	end := s.End.In(timezone).Time()
+
+	message += fmt.Sprintf("\n%s-%s (%s)", start.Format(time.Kitchen), end.Format(time.Kitchen), timezone)
```
mickmister, line 73: I'm curious what you think of using the lowercase `pm` instead of `time.Kitchen`, so it would be `3:04pm` instead of `3:04PM`. This comment made me think about this: https://github.com/mkraft/mattermost-plugin-whentochat/pull/1#issuecomment-609034663

I also think it makes sense to let the user choose the time format they want their event times to show up as. I think we agreed that on this release we assume `AM/PM` format is fine.

server/mscalendar/views/calendar.go
```diff
@@ -64,3 +64,22 @@ func groupEventsByDate(events []*remote.Event) [][]*remote.Event {
...
+	if s.Subject == "" {
+		message += fmt.Sprintf("\nNo subject.")
```
mickmister, line 75: 0/5

```suggestion
		message += fmt.Sprintf("\n(blank)")
```

larkox: Not sure about this. Any comment @aaronrothschild ?

As a bit of context:
- Since an event can be created without a subject, we have to find a way to show that.
- The normal view (right now) would be the following:
```
You have an upcoming event:
4:00pm-5:00pm (Romance Standard Time)
Subject: Some subject
Location: Some location
```
- My proposal would be to have the subject as "No subject", and therefore...
```
You have an upcoming event:
4:00pm-5:00pm (Romance Standard Time)
No Subject.
Location: Some location
```
- Michael proposal would be to have it as `(blank)`, but not sure if with or without the Subject prefix. Therefore:
```
You have an upcoming event:
4:00pm-5:00pm (Romance Standard Time)
(blank)
Location: Some location
```
or
```
You have an upcoming event:
4:00pm-5:00pm (Romance Standard Time)
Subject: (blank)
Location: Some location
```
- As an extra note, when there is no subject, Microsoft shows `(No subject)` in the calendar.

aaronrothschild: @larkox I would go with "unknown" since the user may assume that they accidentally had a subject of "blank"

server/mscalendar/availability.go
larkox: It is one of the multiple workarounds to one weird issue with multiple variable declaration.
I need timezone to be declared outside of the for loop. But I need to give it value inside the `if timezone == ""`.
If I try to assign err inside the if like `timezone, err = ...` will complain about err not being declared. If I do the implicit declaration with `timezone, err := ...` it will also create a new variable timezone that will shadow my previous one.
Since I am going to use the variable err in more places, I thought it ok to declare it before the if, but it would make sense only inside the if.
Another approach would be doing something like `aux, err := ...` and then `timezone = aux`.

server/mscalendar/views/calendar.go
larkox: That could be a wonderful help wanted ticket.

DHaussermann:
Spoke with @mickmister We want to merge this now to be verified in master.  This will aid in making other PRs more testable.

Issue #47, opened 2020-03-06
When accessing the KVStore, use KVCompareAndSet to avoid issues with concurrent access


mickmister:
#### Summary

In order to be tolerant of different HA nodes concurrently accessing the same KVStore value, we should use `KVCompareAndSet` throughout to ensure that the value has not changed during any update operation.

mickmister:
@lieut-data Could we use `cluster.Mutex` to solve this case? There have been a few implementations of solving concurrent KV store access. `KVCompareAndSet` and [`KVModifyAtomic`](https://github.com/mattermost/mattermost-server/pull/13805/files) (which is in progress and does not use `KVCompareAndSet`).

I want to converge on one solution to guard against concurrent access to the KV store. Using `cluster.Mutex` seems practical because we can guard on the application level, and not rely on the database. It will be more certainly atomic and won't require retries.

lieut-data:
@mickmister, eventually, everything falls back to the `KVCompareAndSet` operation under the covers. `cluster.Mutex` can be appropriate for a critical section, but may not be appropriate to safeguard access to lots of different key values (since you'd need a lock for each key value, which is overkill). 

I hadn't seen the `KVModifyAtomic` proposal before, but it will only be correct if it uses `KVCompareAndSet` (or the equivalent `KVSetWithOptions` and the `Atomic` bit). At the end of the day, it should just amount to a loop that retries a a modification until it succeeds (atomically), or the preconditions to make the change in the first place fail (because of a concurrent update).

So my gut is that this ticket is still valid as originally written.

PR #59, opened 2020-03-18
Use KVSetWithOptions with Atomic: true

mickmister:
#### Summary

This PR uses the `Atomic` KVOption to ensure that the kv value used for the operation is up to date upon save.

The atomic saves are used in all contexts where the current value affects the operation being done. For other cases, we are creating new entries in the store, in which case the names of the methods are StoreNewUser etc.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/47

### DIFF
--- a/server/command/daily_summary.go
+++ b/server/command/daily_summary.go
@@ -40,7 +40,7 @@ func (c *Command) dailySummary(parameters ...string) (string, error) {
 
 		return dailySummaryResponse(dsum), nil
 	case "settings":
-		dsum, err := c.MSCalendar.GetDailySummaryUserSettingsForUser(c.user())
+		dsum, err := c.MSCalendar.GetDailySummarySettingsForUser(c.user())
 		if err != nil {
 			return err.Error() + "\nYou may need to configure your daily summary using the commands below.\n" + dailySummaryHelp, nil
 		}
--- a/server/command/daily_summary.go
+++ b/server/command/daily_summary.go
@@ -40,7 +40,7 @@ func (c *Command) dailySummary(parameters ...string) (string, error) {
 
 		return dailySummaryResponse(dsum), nil
 	case "settings":
-		dsum, err := c.MSCalendar.GetDailySummarySettingsForUser(c.user())
+		dsum, err := c.MSCalendar.GetDailySummaryUserSettingsForUser(c.user())
 		if err != nil {
 			return err.Error() + "\nYou may need to configure your daily summary using the commands below.\n" + dailySummaryHelp, nil
 		}
@@ -64,7 +64,7 @@ func (c *Command) dailySummary(parameters ...string) (string, error) {
 	}
 }
 
-func dailySummaryResponse(dsum *store.DailySummarySettings) string {
+func dailySummaryResponse(dsum *store.DailySummaryUserSettings) string {
 	if dsum.PostTime == "" {
 		return "Your daily summary time is not yet configured.\n" + dailySummarySetTimeErrorMessage
 	}
--- a/server/mscalendar/oauth2.go
+++ b/server/mscalendar/oauth2.go
@@ -96,7 +96,7 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
 		return err
 	}
 
-	uid, err := app.Store.LoadMattermostUserId(me.ID)
+	uid, err := app.Store.LoadMattermostUserID(me.ID)
 	if err == nil {
 		user, userErr := app.PluginAPI.GetMattermostUser(uid)
 		if userErr == nil {
@@ -123,9 +123,14 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
 
 	if mattermostUserID == app.Config.BotUserID {
 		app.Poster.DM(authedUserID, BotWelcomeMessage, me.Mail)
-	} else {
-		app.Poster.DM(mattermostUserID, WelcomeMessage, me.Mail)
+		return nil
 	}
 
+	err = app.Store.StoreUserInIndex(u)
+	if err != nil {
+		return err
+	}
+
+	app.Poster.DM(mattermostUserID, WelcomeMessage, me.Mail)
 	return nil
 }
--- a/server/store/daily_summary_store.go
+++ b/server/store/daily_summary_store.go
@@ -4,16 +4,20 @@
 package store
 
 import (
+	"encoding/json"
+
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/kvstore"
 )
 
 type DailySummaryStore interface {
 	LoadDailySummaryIndex() (DailySummaryIndex, error)
-	SaveDailySummaryIndex(dsumIndex DailySummaryIndex) error
-	DeleteDailySummarySettings(mattermostUserID string) error
+	LoadDailySummaryUserSettings(mattermostUserID string) (*DailySummaryUserSettings, error)
+	StoreDailySummaryUserSettings(dsum *DailySummaryUserSettings) error
+	DeleteDailySummaryUserSettings(mattermostUserID string) error
+	ModifyDailySummaryIndex(modify func(dsumIndex DailySummaryIndex) (DailySummaryIndex, error)) error
 }
 
-type DailySummarySettings struct {
+type DailySummaryUserSettings struct {
 	MattermostUserID string `json:"mm_id"`
 	Enable           bool   `json:"enable"`
 	PostTime         string `json:"post_time"` // Kitchen format, i.e. 8:30AM
@@ -21,7 +25,7 @@ type DailySummarySettings struct {
 	LastPostTime     string `json:"last_post_time"`
 }
 
-type DailySummaryIndex []*DailySummarySettings
+type DailySummaryIndex []*DailySummaryUserSettings
 
 func (s *pluginStore) LoadDailySummaryIndex() (DailySummaryIndex, error) {
 	dsumIndex := DailySummaryIndex{}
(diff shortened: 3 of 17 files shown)


mickmister:
There are open PRs that need to be merged, before this PR can include the fixes for the KV store changes made in the PRs:
- [Daily summary job](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/36)
- [Login workflow](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/44)

@larkox Anything in the `Login workflow` PR related to KV store concurrency can be deferred to this PR.

server/utils/kvstore/kvstore.go
```diff
@@ -56,3 +59,46 @@ func StoreJSON(s KVStore, key string, v interface{}) (returnErr error) {
...
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "modification error")
+		}
+
+		return initialBytes, modifiedBytes, nil
+	}
+
+	var (
+		retryLimit     = 5
+		retryWait      = 30 * time.Millisecond
```
larkox: Any strong reason for 30 ms or is it just a magic number?

server/utils/kvstore/kvstore.go
```diff
@@ -56,3 +59,46 @@ func StoreJSON(s KVStore, key string, v interface{}) (returnErr error) {
 	}
 	return s.Store(key, data)
 }
+
+func AtomicModify(s KVStore, key string, modify func(initialValue []byte) ([]byte, error)) error {
```
larkox: Not sure why not including this as a method of the plugin store.

server/mscalendar/oauth2.go
```diff
@@ -123,9 +123,15 @@ func (app *oauth2App) CompleteOAuth2(authedUserID, code, state string) error {
 
 	if mattermostUserID == app.Config.BotUserID {
 		app.Poster.DM(authedUserID, BotWelcomeMessage, me.Mail)
-	} else {
-		app.Poster.DM(mattermostUserID, WelcomeMessage, me.Mail)
+		return nil
 	}
 
+	err = app.Store.StoreUserInIndex(u)
```
larkox: Just to make sure: we do not want the bot user to be in the user index? It seems that we used to have it there (I might be wrong).

larkox: Another potential problem here is... What if I cannot store the index? What happens with the "detached user" (a user stored in the KVStore but not referenced in the index)?

My first thought is that will not be a big problem, since the user will try to connect again, and will overwrite the value, but I wonder what are your thoughts on that.

server/utils/kvstore/kvstore.go
larkox: May be better here, but wondering why not putting this on the top of the file as constants.

mickmister: It's the magic number we used in the Jira plugin

server/utils/kvstore/kvstore.go
mickmister: It is a method of the store, which proxies to this function. This way of doing it allows us to synthetically put the method on the `KVStore interface`, rather than a struct. So any implementation of `KVStore` can "have" this function. This is particularly useful for the `hashedKeyStore`.

server/mscalendar/oauth2.go
mickmister: Yes, I am kind of doing a drive-by refactor that makes it so we don't have to access the `userIndex` twice during the status sync job. The first access is usually spent filtering out the bot user, but if it is not in the index to begin with (which is want we originally wanted to do), then we do not need to filter it out.

mickmister: Hmmm I'm not sure. I believe any operations trying to access the user from the index will gracefully fail (meaning no panic). I think we should tell the user to try again from here. I'm actually not sure what the behavior is to the user when they hit an error here.

mickmister: To be clear, this sort of change would not be acceptable if we had already released the plugin. But since it is still in development, it is okay to change different assumptions about how the store is used.

larkox: My concern is more about leaving "thrash" on the KVStore, but probably not a real problem.

mickmister: Right, almost like we need a `transaction` for multiple KV keys. Using a `cluster.Mutex` and rollback functionality, we can implement it. @levb and I have talked about this. We would need to design it into the `store` or `utils/kvstore` package.

mickmister:
The atomic saves are now used in all contexts where the current value affects the operation being done. For other cases, we are creating new entries in the store, in which case the names of the methods are `StoreNewUser` etc.

server/store/daily_summary_store.go
```diff
@@ -32,25 +35,73 @@ func (s *pluginStore) LoadDailySummaryIndex() (DailySummaryIndex, error) {
...
-	users, err := s.LoadDailySummaryIndex()
-	if err != nil {
-		return err
-	}
+func (s *pluginStore) AtomicModifyUserDailySummarySettings(mattermostUserID string, modify func(settingToStore *DailySummarySettings) (*DailySummarySettings, error)) error {
+	return s.AtomicModifyDailySummaryIndex(func(dsumIndex DailySummaryIndex) (DailySummaryIndex, error) {
+		toChange := &DailySummarySettings{MattermostUserID: mattermostUserID}
+		for _, setting := range dsumIndex {
+			if setting.MattermostUserID == mattermostUserID {
+				toChange = setting
```
larkox: A break here would be useful, right? In theory there should be one only entry for each ID, so after first we can break the loop.

server/store/daily_summary_store.go
```diff
@@ -32,25 +35,73 @@ func (s *pluginStore) LoadDailySummaryIndex() (DailySummaryIndex, error) {
...
+		updated, err := modify(toChange)
+		if err != nil {
+			return nil, err
 		}
-	}
-	return s.SaveDailySummaryIndex(result)
+
+		result := DailySummaryIndex{}
+		found := false
+		for _, setting := range dsumIndex {
```
larkox: In order not to run over the dsumIndex twice, couldn't we store the index from the previous loop?

Also, you can use the index for knowing if it was found or not.

server/store/daily_summary_store.go
```diff
@@ -32,25 +35,73 @@ func (s *pluginStore) LoadDailySummaryIndex() (DailySummaryIndex, error) {
...
+}
+
+func (s *pluginStore) AtomicDeleteUserDailySummarySettings(mattermostUserID string) error {
+	return s.AtomicModifyDailySummaryIndex(func(dsumIndex DailySummaryIndex) (DailySummaryIndex, error) {
+		result := DailySummaryIndex{}
+		for _, setting := range dsumIndex {
+			if setting.MattermostUserID != mattermostUserID {
+				result = append(result, setting)
+			}
+		}
```
larkox: Wouldn't it be better something in this line?
```suggestion
		for i, setting := range dsumIndex {
			if setting.MattermostUserID == mattermostUserID {
				result = append(dsumIndex[:i-1], dsumIndex[i+1:]...)
				break
			}
		}
```

levb:
> ...in all contexts where the current value affects the operation being done.

In other words, for operations that span multiple requests/timers. I am not 100% if I can

mickmister:
@levb Not sure why, but ci is failing on the `TestNormalizeRemoteBaseURL` test

```
--- FAIL: TestNormalizeRemoteBaseURL/[jdsh (0.00s)
        utils_test.go:50: 
            	Error Trace:	utils_test.go:50
            	Error:      	Not equal: 
            	            	expected: "parse //[jdsh: missing ']' in host"
            	            	actual  : "parse \"//[jdsh\": missing ']' in host"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-parse //[jdsh: missing ']' in host
            	            	+parse "//[jdsh": missing ']' in host
            	Test:       	TestNormalizeRemoteBaseURL/[jdsh
```

hanzei:
@mickmister https://github.com/mattermost/mattermost-plugin-mscalendar/pull/63 will fix the build

server/mscalendar/subscription.go
```diff
@@ -81,17 +81,16 @@ func (m *mscalendar) RenewMyEventSubscription() (*store.Subscription, error) {
 		return nil, err
 	}
 
-	storedSub, err := m.Store.LoadSubscription(m.actingUser.Settings.EventSubscriptionID)
-	if err != nil {
-		return nil, err
+	var subToReturn *store.Subscription
```
larkox: Maybe I am missing something, but I do no see where subToReturn is assigned any value.

server/store/daily_summary_store.go
```diff
@@ -64,44 +64,40 @@ func (s *pluginStore) AtomicModifyDailySummaryIndex(modify func(dsumIndex DailyS
...
-		for _, setting := range dsumIndex {
-			if setting.MattermostUserID != mattermostUserID {
-				result = append(result, setting)
+		for i, setting := range dsumIndex {
+			if setting.MattermostUserID == mattermostUserID {
+				return append(dsumIndex[:i], dsumIndex[i+1:]...), nil
 			}
 		}
-		return result, nil
+		return dsumIndex, nil
```
larkox: 0/5 Should return an error when it cannot find the element to delete? Probably not, but just wondering.

server/store/subscription_store.go
```diff
@@ -76,3 +84,30 @@ func (s *pluginStore) DeleteUserSubscription(user *User, subscriptionID string)
...
+
+		var sub *Subscription
+		if len(initial) > 0 {
+			err := json.Unmarshal(initial, sub)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		updated, err := modify(sub)
```
larkox: If len(initial) == 0, here sub is null. Is this the expected behaviour?

server/store/user_store.go
```diff
@@ -157,6 +169,24 @@ func (s *pluginStore) AtomicStoreUserInIndex(user *User) error {
...
+			Email:            user.Remote.Mail,
+		}
+
+		for i, u := range userIndex {
+			if u.MattermostUserID == user.MattermostUserID && u.RemoteID == user.Remote.ID {
+				result := append(userIndex[:i], newUser)
+				return append(result, userIndex[i+1:]...), nil
+			}
+		}
+		return userIndex, nil
```
larkox: 0/5 should return an error if it cannot find the user?

mickmister: This is a bug. The user should be appended to the end of the list if not found in the loop.

server/store/subscription_store.go
mickmister: If `sub == nil` then the error at `line 90` would have returned already because of not finding it in the store. The `if len(initial) > 0` is a bit redundant. What do you suggest here, an additional check for `nil` just in case?

server/store/daily_summary_store.go
mickmister: This depends on whether we store the bot in the indexes. In the `store` package we do not have access to the config, so we can't assert whether the user is the bot or not. If we don't find the user for whatever reason, I don't think we would want to cancel the caller's operation anyway.

mickmister: This is really something that we need to definitively understand and agree on though. The implementation detail of not having the bot in the indexes, versus having to check in different places if the user we are interacting with is the bot.

server/store/subscription_store.go
larkox: If I understand correctly, len(initial) should be always >0 because if not, it would return a storeErr, right? If that is the case, maybe you can drop the if statement, and just leave a comment like `initial will always have data`. What do you think?

server/store/daily_summary_store.go
larkox: Does it make sense to delete the daily summary of a bot user? If the bot user should never get a daily summary, and something tries to delete its daily summary, we would consider that an error, am I right?

Nevertheless, I also wonder on concurrency cases. If we try to delete an index that does not exist, is it an error? Or we can safely assume that the action that the user wanted to happen is done (that index is no longer there)?

mickmister: > Does it make sense to delete the daily summary of a bot user?

Depending on how we initialize the daily summary settings (right now they are not automatically initialized on connect), the bot may not be in the daily summary settings.

> Nevertheless, I also wonder on concurrency cases. If we try to delete an index that does not exist, is it an error? Or we can safely assume that the action that the user wanted to happen is done (that index is no longer there)?

As of now, this function is only called when a user disconnects, including when we disconnect the bot. If deletion in the index doesn't happen, the most we should do is log it. However, I wouldn't want it to log if it is the bot user.

We could make it so the `DisconnectUser` function that calls this function checks if it is the bot user, and only calls the relevant store methods. This would allow us to correctly return and log errors for real users.

server/utils/kvstore/kvstore.go
```diff
@@ -56,3 +65,54 @@ func StoreJSON(s KVStore, key string, v interface{}) (returnErr error) {
 	}
 	return s.Store(key, data)
 }
+
+func AtomicModifyWithOptions(s KVStore, key string, modify func(initialValue []byte, storeErr error) ([]byte, *model.PluginKVSetOptions, error)) error {
+	readModify := func() ([]byte, []byte, *model.PluginKVSetOptions, error) {
```
levb: 1/5 what are the benefits of this func? It is only used one, I'd inline it.

server/store/user_store.go
```diff
@@ -12,11 +13,15 @@ import (
...
-	LoadMattermostUserId(remoteUserId string) (string, error)
+	LoadUser(mattermostUserID string) (*User, error)
+	LoadMattermostUserID(remoteUserID string) (string, error)
 	LoadUserIndex() (UserIndex, error)
-	StoreUser(user *User) error
-	DeleteUser(mattermostUserId string) error
+	StoreNewUser(user *User) error
+	LoadUserFromIndex(mattermostUserID string) (*UserShort, error)
+	DeleteUser(mattermostUserID string) error
+	AtomicModifyUserIndex(modify func(userIndex UserIndex) (UserIndex, error)) error
```
levb: Why use `Atomic...` prefix in the names if there are no non-atomic equivalents to disambiguate from?

mickmister:
@levb and I spoke offline and decided its not worth the complexity to atomically modify single-record key values. I've resubmitted my code here with just the multi-record values being atomically modified.

cc @larkox

server/command/daily_summary.go
```diff
@@ -40,7 +40,7 @@ func (c *Command) dailySummary(parameters ...string) (string, error) {
 
 		return dailySummaryResponse(dsum), nil
 	case "settings":
-		dsum, err := c.MSCalendar.GetDailySummarySettingsForUser(c.user())
+		dsum, err := c.MSCalendar.GetDailySummaryUserSettingsForUser(c.user())
```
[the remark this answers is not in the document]
mickmister: > Maybe it can be left like this for parallelism with the other object to return

@larkox I'm not sure I understand. Can you elaborate on this?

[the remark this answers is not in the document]
larkox: The class of the object to return is called (if I remember correctly) DailySummaryUserSettings. Therefore, you are getting a "DailySummaryUserSettings" for one user. Nevertheless, it feels redundant.

levb:
(leaving it as "Dev Review for @mickmister to reconsider one last time before it's merged :) )

DHaussermann:
Spoke with @mickmister. As testing this would require end to end regression testing, it would be best to test post merge in master along with other code changes.

Issue #61, opened 2020-03-18
Settings panel


larkox:
Create a Settings Panel where the user can change the user settings.

PR #62, opened 2020-03-18
Add settings panel

larkox:
#### Summary
Created an implementation for the settings panel which use post messages with slack attachments. This setting panel should be reusable for other projects.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/61


larkox:
Current implementation look like this:
image

Also allow disabling some settings based on the status of others:
image

aaronrothschild:
Hmmm.  Is there a way to indicate via the interactive buttons what is currently set?  (ie: an active border around "Yes") 

@asaadmahmood any ideas on how to best articulate to the user what is currently set already?

> Current implementation look like this:
> image

I'm thinking of changing the verbiage on current status to: 
- "Currently, I will send you an update before updating your status" or 
- "Currently, I will not send you an update before updating your status"

@asaadmahmood Should we use "I" to refer to the CalendarBot itself?



> Also allow disabling some settings based on the status of others:
> image

larkox:
If we were to wait for [this PR](https://github.com/mattermost/mattermost-webapp/pull/4993) we would be able to apply some color to the one selected. Not sure if it is the best way to show it.

asaadmahmood:
I didn't get this, what is "I"?
"Should we use "I" to refer to the CalendarBot itself?" 

For the second one, I'd suggest:
Do you want to update your status to **DND** in Mattermost when you are in a meeting?
**Yes**  **No**

I'm guessing the user only sees this one, not again and again, because if he has selected either yes or no, that's his setting, and he can change it when he wants, rather than seeing that message popup time and time again.

So it makes sense not to focus on what the setting is right now i.e. Currently, and just ask him what he wants to do.

larkox:
@asaadmahmood with "I" Aaron refers to "I" as in "I, me, myself" or "I, you, he".

The idea of this is to work as a "panel". It will be always present there until it is somehow closed (in the current implementation, it is only when we disconnect, or write /settings again). It would be a similar idea to go to the system settings, and change settings there.

asaadmahmood:
Got it. Yeah I think we can use "I" to get a more personal feel.
This post wouldn't be used much. Once I press yes or no, after a while, it may even scroll above, and I may need to access my settings from elsewhere to reconfigure the status settings.

larkox:
You are right. The post will be seldom reused, and we will have to write `/mscalendar settings` every time we want to change a setting, since the post will scroll above, but I kind of like the idea of persistent panel. That way, you also are able to better see dependency between settings.
Finally, knowing the current state is important, since you are not changing here one single setting. You are taking a look at all the settings, and changing the one you want.

server/utils/settingspanel/handler.go
```diff
@@ -0,0 +1,53 @@
+package settingspanel
+
+import (
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/httputils"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+type sh struct {
```
levb: 1/5 s/sh/handler/

server/utils/bot/poster.go
```diff
@@ -16,45 +16,56 @@ type Poster interface {
...
+	DMWithAttachments(mattermostUserID string, attachments ...*model.SlackAttachment) (string, error)
 
 	// Ephemeral sends an ephemeral message to a user
 	Ephemeral(mattermostUserID, channelID, format string, args ...interface{})
+
+	// DeletePost deletes a single post
+	DeletePost(postID string) error
+
+	// DMUpdatePost substitute one post with another
+	DMUpdatePost(post *model.Post) error
```
levb: s/DMUpdatePost/UpdatePost/ ??

levb:
Nice stuff! I think it'll need to mature and get simplified a bit, but I want to use it, now!

server/store/setting_store.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/kvstore"
+)
+
+const (
+	UpdateStatusSettingID    = "update_status"
+	GetConfirmationSettingID = "get_confirmation"
+)
+
+func (s *pluginStore) SetSetting(userID, settingID string, value interface{}) error {
```
mickmister: `value interface{}` Great, I see this can support strings as well :+1:

server/plugin/plugin.go
```diff
@@ -133,7 +133,17 @@ func (p *Plugin) OnConfigurationChange() (err error) {
...
 		e.Dependencies.Store = store.NewPluginStore(p.API, e.bot)
 		e.Dependencies.IsAuthorizedAdmin = p.IsAuthorizedAdmin
-		e.Dependencies.SettingsPanel = mscalendar.NewSettingsPanel(e.bot, e.Dependencies.Store, e.Dependencies.Store, "/settings", pluginURL)
+		e.Dependencies.SettingsPanel = mscalendar.NewSettingsPanel(
+			e.bot,
+			e.Dependencies.Store,
+			e.Dependencies.Store,
+			"/settings",
+			pluginURL,
+			func(userID string) string {
```
mickmister, line 142: I think we should return `(string, error)`. As I currently understand, if we can't get their timezone, then we can't properly set the settings.

server/mscalendar/settings_daily_summary.go
```diff
@@ -0,0 +1,245 @@
...
+				},
+			},
+			Type:          "select",
+			Options:       s.makeAPMOptions(currentH, currentM),
+			DefaultOption: fullTime,
+		}
+
+		actions = []*model.PostAction{&actionOptionsH, &actionOptionsM, &actionOptionsAPM}
+
+		if currentTextValue != "Not set" {
```
mickmister, line 167: `currentTextValue` is initialized as `Not set.` but is being compared to `Not set` here.

server/store/setting_store.go
```diff
@@ -61,22 +63,51 @@ func (s *pluginStore) GetSetting(userID, settingID string) (interface{}, error)
 	}
 }
 
-func (s *pluginStore) getDailySummarySettingForUser(userID string) (string, error) {
+func (s *pluginStore) getDailySummarySettingForUser(userID string) (*DailySummarySettings, error) {
```
mickmister, line 67: `LoadDailySummaryUserSettings` is available on master

server/store/setting_store.go
```diff
@@ -84,24 +85,46 @@ func (s *pluginStore) updateDailySummarySettingForUser(userID string, value inte
...
 
-			if stringValue == "false" {
+			if splittedValue[0] == "false" {
 				dsum.Enable = false
 				break
 			}
 
-			dsum.PostTime = value.(string)
+			dsum.PostTime = splittedValue[0]
+			dsum.Timezone = timezone
```
mickmister: This seems pretty hacky to me. What is the reason for needing to accept `interface{}` here?

Maybe make a struct to pass in here instead of doing string parsing. We can still accept `interface{}`, but assert it to be what we expect it to be. This way, the caller can be more specific with what it's passing.

server/store/setting_store.go
```diff
@@ -61,22 +63,51 @@ func (s *pluginStore) GetSetting(userID, settingID string) (interface{}, error)
...
+				break
+			}
+
+			dsum.PostTime = value.(string)
+			break
 		}
 	}
 
-	return "Daily summary not set", nil
+	err = s.SaveDailySummaryIndex(dsumIndex)
```
mickmister, line 128: `StoreDailySummaryUserSettings` is available so you don't need to access the index directly

server/utils/settingspanel/handler.go
```diff
@@ -46,8 +46,10 @@ func (sh *handler) handleAction(w http.ResponseWriter, r *http.Request) {
 	sh.panel.Set(mattermostUserID, id, value)
 
 	response := model.PostActionIntegrationResponse{}
+	post, err := sh.panel.GetUpdatePost(mattermostUserID)
```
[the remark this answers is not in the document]
larkox: I think there is no need to be verbose with `settings panel`, since it is already the object we are working with.
Maybe NewPost is clear enough?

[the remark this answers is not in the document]
larkox: After a second though, I feel it makes more sense "ToPost", since we are converting the panel into a post.

server/mscalendar/settings_daily_summary.go
```diff
@@ -0,0 +1,245 @@
...
+		os.optionsM = append(os.optionsM, fmt.Sprintf("%02d", i*15))
+	}
+
+	os.optionsAPM = []string{"AM", "PM"}
+
+	return os
+}
+
+func (s *dailySummarySetting) Set(userID string, value interface{}) error {
+	err := s.store.SetSetting(userID, s.id, value)
```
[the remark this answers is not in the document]
larkox: If we were to trust the program invariant we could remove the one from Get, actually. Nevertheless, it seems a sensible thing to do.
Nevertheless, note that here we are expecting a string here, not a DailySummarySetting. The Store is the one in charge of making that string into a DailySummarySetting.

server/store/setting_store.go
larkox: There are many considerations here to do it this way.

First of all, this function is called (with a few steps) by the handler, which is agnostic.
The handler passes the context in case of buttons, the option in case of selects. Since this is a select in general, we should receive a string, which is the only value an option can have, and then parse it.

Regarding receiving interface instead of string, is just to make the type conversions where it makes sense, which is here. Otherwise, we would have to make the conversions in the case block on the switch on SetSetting.

We could put all the parsing in one function that returns this struct you are talking about, but I am not sure whether it would improve the readibility.

mickmister: What happens when this value gets more complicated, or we have a more complicated value we have to handle similarly? Is there anyway the origin of the data can tell this function "this is the post time", or "this is the enable flag"? For example, if this store value had more than one boolean, this would not work.

server/mscalendar/settings_daily_summary.go
```diff
@@ -0,0 +1,259 @@
...
+func (s *dailySummarySetting) GetDependency() string {
+	return s.dependsOn
+}
+
+func (s *dailySummarySetting) GetSlackAttachments(userID, settingHandler string, disabled bool) (*model.SlackAttachment, error) {
+	title := fmt.Sprintf("Setting: %s", s.title)
+	currentValueMessage := "Disabled"
+
+	actions := []*model.PostAction{}
+	if !disabled {
```
mickmister, line 97: This `!disabled` block is pretty long, leading up to just about the end of the function. Do you think we can instead have a short `if disabled` block that returns early on?

server/mscalendar/settings_daily_summary.go
```diff
@@ -0,0 +1,259 @@
...
+			}
+			currentTextValue = fmt.Sprintf("%s (%s) (%s)", dsum.PostTime, dsum.Timezone, enableText)
+		}
+
+		timezone, err := s.getTimezone(userID)
+		if err != nil {
+			if dsum != nil {
+				timezone = dsum.Timezone
+			} else {
+				timezone = "UTC"
```
mickmister, line 130: I'm not sure what benefit the user gets in setting their settings with the wrong timezone. 1/5 Instead of falling back on UTC, return an error here saying that we couldn't get their timezone from Microsoft.

server/mscalendar/settings_daily_summary.go
```diff
@@ -0,0 +1,259 @@
...
+				},
+			},
+			Type:          "select",
+			Options:       s.makeAPMOptions(currentH, currentM, timezone),
+			DefaultOption: fullTime,
+		}
+
+		actions = []*model.PostAction{&actionOptionsH, &actionOptionsM, &actionOptionsAPM}
+
+		buttonText := "Enable"
```
mickmister, line 178: What purpose did the `if currentTextValue == "Not set."` block serve that is no longer necessary? Just want to understand the code.

server/store/setting_store.go
larkox: We could codify the option content as a JSON string. What do you think?

server/mscalendar/settings_daily_summary.go
larkox: The idea was not to show the Enable button before a Daily Summary was set, but I thought that it would make sense to allow the enable with the default value.

aaronrothschild:
LGTM

PR #63, opened 2020-03-28
Fix build

hanzei:
#### Summary
- Fix test under go 1.14
- Update package-lock.json

### DIFF
--- a/server/utils/httputils/utils_test.go
+++ b/server/utils/httputils/utils_test.go
@@ -28,7 +28,7 @@ func TestNormalizeRemoteBaseURL(t *testing.T) {
 
 		// Errors
 		{"[jdsh", "", "",
-			`parse //[jdsh: missing ']' in host`},
+			`parse "//[jdsh": missing ']' in host`},
 		{"/mmtest", "", "",
 			`Invalid URL, no hostname: "/mmtest"`},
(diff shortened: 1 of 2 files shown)


server/utils/httputils/utils_test.go
```diff
@@ -28,15 +28,15 @@ func TestNormalizeRemoteBaseURL(t *testing.T) {
...
+			`parse "//[jdsh": missing ']' in host`},
 		{"/mmtest", "", "",
 			`Invalid URL, no hostname: "/mmtest"`},
 		{"/mmtest/", "", "",
 			`Invalid URL, no hostname: "/mmtest/"`},
 		{"http:/mmtest/", "", "",
 			`Invalid URL, no hostname: "http:/mmtest/"`},
 		{"hƒƒp://xyz.com", "", "",
-			`parse hƒƒp://xyz.com: first path segment in URL cannot contain colon`},
+			`parse "hƒƒp://xyz.com": first path segment in URL cannot contain colon`},
```
mickmister: What are these ƒƒ characters? It seems like the ƒƒ is the only reason that this test passes, because otherwise the url is valid.

hanzei: I have no idea :man_shrugging:

PR #72, opened 2020-04-02
Use batch requests for daily summary job

### LINKED ISSUE #48: Use batch requests in the Daily Summary job to improve performance
#### Summary

As of https://github.com/mattermost/mattermost-plugin-mscalendar/pull/36, the daily summary job does a separate http request for each user to gather the calendar data.

If we instead queue up which users need their daily summaries "now", then use [batch requests](https://docs.microsoft.com/en-us/graph/json-batching) to send them in a 20-request batch, this will tremendously improve the performance. You can find example code using batch requests in the [`GetSchedule`](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/864a161240b2cb28b3496c1dcc45a16204f23e54/server/remote/msgraph/get_schedule_batched.go#L50) method in the `remote/msgraph` package.

mickmister:
#### Summary

The daily summary job now uses msgraph JSON batching to limit the number of requests made. Order of operations:

- Fetch DailySummaryIndex from store
- Filter out users who should not receive their summary
- Perform API requests with 20 users per request
- Send each user their summary

#### Ticket Link

Fixes #48


server/mscalendar/daily_summary.go
```diff
@@ -120,86 +131,94 @@ func (m *mscalendar) SetDailySummaryEnabled(user *User, enable bool) (*store.Dai
...
 	return nil
 }
 
-func (m *mscalendar) getDailySummary(user *User) (string, error) {
-	tz, err := m.GetTimezone(user)
-	if err != nil {
-		return "", err
-	}
+func (m *mscalendar) GetDailySummaryForUser(user *User) (string, error) {
+	return m.getDailySummaryForUser(user)
```
larkox, line 203: Why using a private function? Couldn't we add this functionality directly here?

server/mscalendar/daily_summary.go
```diff
@@ -120,86 +131,94 @@ func (m *mscalendar) SetDailySummaryEnabled(user *User, enable bool) (*store.Dai
...
 
-	if len(calendarData) == 0 {
-		m.Poster.DM(user.MattermostUserID, "You have no upcoming events today.")
-		return "", nil
-	}
-
-	postStr, err := views.RenderCalendarView(calendarData, tz)
+	calendarData, err := m.getTodayCalendarEvents(user, time.Now(), tz)
 	if err != nil {
+		m.Poster.DM(user.MattermostUserID, "Failed to get user . %s", err.Error())
```
larkox, line 214: Should this function do the error posting to the user, or should that be handled outside?

mickmister: This is very strange, that's not even the right error message. How about:

```go
return "Failed to get calendar events", err
```

The string return value gets sent back to the user for the slash command response.

server/mscalendar/daily_summary.go
mickmister: Yes, this function was used elsewhere but now it is not :+1:

server/mscalendar/daily_summary_job.go
```diff
@@ -74,7 +74,7 @@ func (job *DailySummaryJob) Cancel() {
 func (j *DailySummaryJob) work() {
 	j.Logger.Debugf("Daily summary job beginning")
 
-	err := New(j.Env, "").ProcessAllDailySummary()
+	err := New(j.Env, "").ProcessAllDailySummary(time.Now())
```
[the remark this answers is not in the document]
mickmister: It makes the function more testable by injecting the dependency of `now`.

server/command/daily_summary.go
```diff
@@ -59,6 +60,12 @@ func (c *Command) dailySummary(parameters ...string) (string, error) {
 			return err.Error(), err
 		}
 		return dailySummaryResponse(dsum), nil
+	case "all":
```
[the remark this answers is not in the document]
mickmister: This is for debugging purposes. The functionality is limited to admins. I can't see it being useful to admins, as it is essentially "Run this job now even though it will run in the next 10-15 minutes." Also, it would only run for users that are in the correct time window anyway. Maybe just take it out?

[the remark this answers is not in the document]
mickmister: Removed this block

levb:
@mickmister Can you please resolve the conflict?

server/jobs/daily_summary_job.go
```diff
@@ -25,16 +26,9 @@ func NewDailySummaryJob() RegisteredJob {
...
-	_, err := mscal.GetRemoteUser(env.BotUserID)
-	if err != nil {
-		env.Logger.Errorf("Please connect bot user using `/%s connect_bot`, in order to run the daily summary job.", config.CommandTrigger)
-		return
-	}
-
 	env.Logger.Debugf("Daily summary job beginning")
 
-	err = mscal.ProcessAllDailySummary()
+	err := mscalendar.New(env, "").ProcessAllDailySummary(time.Now())
```
levb, line 31: nit: should this be `BotUserID`?

mickmister: @levb Yes it is implicitly `BotUserID`. The `New` function falls back to the `BotUserID` when an empty string is passed in. It's convenient but I think I prefer the explicit way. Leaving it as is, as we have done this elsewhere and we this is already QA reviewed etc

Issue #60, opened 2020-03-18
Confirmation of a Status Change upcoming


larkox:
The plugin can change your Mattermost status when you are about to enter a meeting. According to the settings, this will be done automatically, or will ask for your confirmation before doing it.

This issue includes adding the logic to get this confirmation from the user.

PR #73, opened 2020-04-03
Add status change confirmation

larkox:
#### Summary
Add confirmation for status change. Whenever the status is about to change, if the user has it configured for so, he will receive a message to confirm whether he wants to change the status.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-mscalendar/issues/60


larkox:
There is a big concern regarding the handling of the availability. Since there is no reliable way to distinguish events during the status sync, we are storing the last availability value we found. If the value has not changed, it does not try to update the status. In general this will work fine, but there are some special cases where it will not work. The most prominent I found is the following:

- A meeting finishes soon, so I change back my status to online. Nevertheless, since the sync status job keeps getting a availability value of busy, I will not get asked to change the status to dnd on the next meeting.

@aaronrothschild How big do you think is this problem?
@levb @mickmister Any idea on how to deal better with this?

larkox:
Tests are failing due to the changes on how we handle the status sync. When all the pieces are there (user config) and the way to avoid asking many times for the user are decided, I will update the tests.

server/mscalendar/availability.go
```diff
@@ -138,24 +140,53 @@ func (m *mscalendar) GetAvailabilities(remoteUserID string, scheduleIDs []string
...
 			return fmt.Sprintf("User is free. Setting user from %s to online.", currentStatus)
 		} else {
 			return fmt.Sprintf("User is free, and is already set to %s.", currentStatus)
 		}
 	case remote.AvailabilityViewTentative, remote.AvailabilityViewBusy:
 		if currentStatus != "dnd" {
+			if m.ShouldNotifyStatusChange(u) {
+				userQuestion := fmt.Sprintf("Shall we change your status from %s to dnd?", currentStatus)
+				m.Poster.DMWithAttachments(mattermostUserID, statusChangeAttachments(userQuestion, "dnd", url))
+				return fmt.Sprintf("User is busy. Setting user from %s to dnd.", currentStatus)
```
mickmister, line 176:
```suggestion
				return fmt.Sprintf("User asked for status change from %s to dnd", currentStatus)
```

server/mscalendar/availability.go
```diff
@@ -138,24 +140,53 @@ func (m *mscalendar) GetAvailabilities(remoteUserID string, scheduleIDs []string
...
 			return fmt.Sprintf("User is busy. Setting user from %s to dnd.", currentStatus)
 		} else {
 			return fmt.Sprintf("User is busy, and is already set to %s.", currentStatus)
 		}
 	case remote.AvailabilityViewOutOfOffice:
 		if currentStatus != "offline" {
+			if m.ShouldNotifyStatusChange(u) {
+				userQuestion := fmt.Sprintf("Shall we change your status from %s to offline?", currentStatus)
+				m.Poster.DMWithAttachments(mattermostUserID, statusChangeAttachments(userQuestion, "offline", url))
+				return fmt.Sprintf("User is out of office. Setting user from %s to offline", currentStatus)
```
mickmister, line 188:
```suggestion
				reeturn fmt.Sprintf("User asked for status change from %s to offline", currentStatus)
```

server/api/post_action.go
```diff
@@ -81,3 +82,49 @@ func (api *api) postActionRespond(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	returnText := "The status has not been changed."
+	if value {
+		changeTo, ok := request.Context["change_to"]
+		if !ok {
+			http.Error(w, "No state to change", http.StatusBadRequest)
+			return
+		}
+		stringChangeTo := changeTo.(string)
```
mickmister, line 115: Nit: can you change this to match the type assertion strategy above?

larkox: Sorry, forgot about that.

server/mscalendar/availability.go
```diff
@@ -138,24 +140,53 @@ func (m *mscalendar) GetAvailabilities(remoteUserID string, scheduleIDs []string
...
+
+	if !m.ShouldChangeStatus(u) {
+		return "Status not changed by user configuration."
+	}
+
+	if !m.HasAvailabilityChanged(u, currentAvailability) {
+		return "Status not changed because there is no update since last status change."
+	}
+
+	u.LastStatusUpdateAvailability = currentAvailability
```
mickmister, line 153: I think we need to store more information here to take into account for the user manually changing their status. Here's a proposal

When the user manually changes their status, we take all events that overlap with `now`, and store them as hashes `(eventid-starttime)`. Those events are now "immune" to an availability change. If an event's start gets moved, it becomes a "new" event as far as the availability changer is concerned.

I think this solves the one-off issue when the user wants to cancel out their most recent status change. If they want it to remain untouched by the automation, then they can disable it in their settings.

larkox: One of the main problems is that there is no eventID that I can use to handle this, I can only work with times and availability.

I can do the following:
- Check if the availability has changed. If changed:
  - Check that mattermost status and handle it
- If the availability is not changed, check the latest start time registered, and compare it with the elements returned.
- If there is a new event (an event that starts later than the last registered) check with the current mattermost status.
- If no event starts later, we assume it is the same event and ignore it.

I will implement this and keep thinking about it. Probably in code make more sense.

mickmister: Hmm well if we don't have an event id we're kind of at a disadvantage in general. Maybe we shouldn't even be using `GetSchedule` since we aren't getting the performance we were getting before with it, which was looking at several users' schedules in one request, the main reason why we had the dedicated bot account.

`GetSchedule`'s main purpose is, find some time where one or more of these people are free given some time constraints. That's not really what we're doing here. We just want to know if they're free "right now". We might want to use `GetCalendarView` with a small time window for this instead.

@larkox Let me know what you think about the shortcomings of using `GetSchedule`, versus the alternative `GetCalendarView`. Which reminds me that we have a `GetSchedule` response json example in the `server/testdata` folder, but not one for `GetCalendarView`.

server/mscalendar/availability.go
```diff
@@ -138,16 +138,20 @@ func (m *mscalendar) GetAvailabilities(remoteUserID string, scheduleIDs []string
...
+	currentAvailability := s.AvailabilityView[0]
 	u := NewUser(mattermostUserID)
 
 	if !m.ShouldChangeStatus(u) {
 		return "Status not changed by user configuration."
 	}
 
 	if !m.HasAvailabilityChanged(u, currentAvailability) {
-		return "Status not changed because there is no update since last status change."
+		ok, newEventTime := m.HasNewEventStarted(u, s)
```
mickmister, line 132: A little verbose, but more obvious

```suggestion
		hasStarted, newEventTime := m.HasNewEventStarted(u, s)
```

server/remote/schedule.go
```diff
@@ -29,7 +38,16 @@ type ScheduleInformation struct {
...
 	// WorkingHours interface{} `json:"workingHours,omitempty"`
 	// Error *FreeBusyError `json:"error,omitempty"`
 }
+
+type ScheduleItem struct {
+	IsPrivate bool
+	Status    string
+	Subject   string
+	Location  string
+	Start     DateTime
```
mickmister, line 51: To follow suit with `Event` use a pointer here, though I mistakenly suggested to not use a pointer earlier.

```suggestion
	Start     *DateTime
```

server/store/user_store.go
```diff
@@ -34,6 +34,7 @@ type User struct {
 	OAuth2Token                  *oauth2.Token
 	Settings                     Settings `json:"mattermostSettings,omitempty"`
 	LastStatusUpdateAvailability byte
+	LastStatusUpdateEventTime    remote.DateTime
```
mickmister, line 42: 1/5 Use a pointer here as well. It keeps the code more uniform and consistent.

mickmister:
@larkox I've updated the logic to take into account events that have already been processed. It should reflect our plans of:

> Store a list of IDs for notified events. If the event is not on the calendar view, remove the id from the list and handle the "back to online". If an event on the calendar view is not in the list, handle the status change. If the list gets emptied, handle the "back to online" situation.

I have not fixed the tests, and there is some potential performance improvements for not doing all `KVGet`s at once, but that can be a follow up discussion after we have a working state of the code in master.

btw the ask for confirmation works great! Nice job :+1:

mickmister:
@larkox The following items have been completed:

- Deletion of dead code in `server/mscalendar/user.go`
- Asserting user is not DND before setting them to DND in the case where I had missed before
- Change pretty status `DND` to `Do Not Disturb`
- Add tests to reflect new calendar event-driven functionality

larkox:
@mickmister Your changes look good to me. Shall we move to QA Review?

DHaussermann:
@larkox I'm not able to test this change because the status sync job no longer seems to be working. 
Since the status sync is not working I can't verify the confirmation on change. Any chance something in this PR affected the status sync? 
Here are my steps:
- Connect a user with MS calendar plugin
- Deploy form this branch
- Login to MS Calendar as mapped user and book an appointment in MS Calendar beginning in a dew minutes
- Confirm appointment is visible with `/mscalendar viewcal`
- Wait until event start time and note that status in Mattermost remains online
cc
@mickmister  I notice for some of the test users that their "IM Status" visible in calendar does not flip to busy during meeting times. Could this affect the sync job or is it unrelated. As in is the sync job purely looking at events and not the user status in outlook.

larkox:
@DHaussermann The latest change made by @mickmister should solve the problem with the status sync working. If you could take another look it would be great :D

mickmister:
@larkox The remaining issue is that we cannot change the `UpdateStatus` and `GetConfirmation` user settings to test this PR. What is the process of:
- Adding these two settings to the settings panel
- Accessing the settings panel with the current code in this PR. This PR is up to date with master which includes the settings panel PR

I'd like to perform these items so I can be more familiar with how to edit the settings panel

larkox:
@DHaussermann This should be ready for another round of QA when you have the time :D

DHaussermann:
@larkox I tried testing this again late yesterday. My problem is not with the sync service or anything related to Mattermost. On MS Calendar side, my user's will not update to busy during an event. I'm not sure why. Adjusting "IM status" manually does not work. I'll look again for some configuration on theMS side that could be affecting this. 
cc @mickmister

larkox:
@DHaussermann Currently, the plugin only updates mattermost status. It does not update anything on Microsoft side (this would be something with a completely different scope). Whether Microsoft changes anything in their IM tool or not should be not important for us, but if there is any configuration about that, we might want in the future consider it.

larkox:
@DHaussermann 
1. The sync service does not look at the status from MS. It just check the events in the next ten minutes (if I am not mistaken) and check if these are "busy events". If they are, ask for status change. We do not check any information about status on MS IM system, or anything similar.

2. `/mscalendar availability` should run the same logic as the sync service for the current user. Therefore, if the settings are set to change automatically the status, it should change automatically, and if set to set confirmation, it should ask for a confirmation. In order to set these settings you can use `/mscalendar settings`.

DHaussermann:
Thanks for the info @larkox much appreciated!  
I will test this again to see if I can get the sync working and will no concern myself with what's happening the the user's status on the MS side.

DHaussermann:
Sorry @larkox  I'm still unable to this feature due to a bug with console.

I cannot set the option for change status and confirm change status in the settings. The buttons are non-responsive and return a 404. 
SettingsButtons

I poked around but the behavior suggests that the setting is false by default when you connect. So, I don't know how to test this without fixing the issue on the login flow branch and pulling it in. 

Just a thought - Is there a way to set the value manually in the DB or something for now so I can test the feature?

mickmister:
In order to properly support status sync, we need to use `GetSchedule` to factor in the user's actual status in Microsoft Calendar. The event notifying code still needs to use the events from `GetCalendarView`, so the logic needs to be separated between the two operations. I'm working on this now and will update this PR to reflect this asap.

larkox:
@mickmister Not sure what you are trying to solve here. Your changes should bring back the problem of several notifications for the same event, and to solve that, the "ugly" handling of events. After a quick look, `GetSchedule` does not even take in mind the user status, just the events and their types.

After some more look into this, it seems that the "presence" status is not even available on MSGraph. There is something on the beta version of the API, but not in the official one.

DHaussermann:
This has ongoing discussion regarding the design.

server/mscalendar/availability.go
```diff
@@ -171,15 +176,20 @@ func (m *mscalendar) setStatusFromAvailability(user *store.User, currentStatus s
 		if e.IsCancelled {
 			continue
 		}
-		h := fmt.Sprintf("%s %s", e.ID, e.Start.Time().UTC().Format(time.RFC3339))
+		h := fmt.Sprintf("%s %s", e.ICalUID, e.Start.Time().UTC().Format(time.RFC3339))
```
mickmister, line 214: Docs on `ICalID`: A unique identifier that is shared by all instances of an event across different calendars. Read-only.

I'm guessing the issues with the `ID` for a given event changing for whatever reason is solved by using `ICalID` instead. Do we want to replace all uses of `ID` in the codebase to avoid the issue elsewhere?

server/jobs/status_sync_job.go
```diff
@@ -15,16 +15,16 @@ func NewStatusSyncJob() RegisteredJob {
 	return RegisteredJob{
 		id:                statusSyncJobID,
 		interval:          mscalendar.StatusSyncJobInterval,
-		work:              runStatusSyncJob,
+		work:              runSyncJob,
 		isEnabledByConfig: isStatusSyncJobEnabled,
 	}
 }
 
 // runStatusSyncJob synchronizes all users' statuses between mscalendar and Mattermost.
```
mickmister, line 23:
```suggestion
// runSyncJob synchronizes all users' statuses between mscalendar and Mattermost.
```

server/mscalendar/availability.go
```diff
@@ -90,18 +90,56 @@ func (m *mscalendar) syncStatusUsers(userIndex store.UserIndex) (string, error)
...
+		if !ok {
+			continue
+		}
+		if s.Error != nil {
+			m.Logger.Errorf("Error getting availability for %s: %s", user.Remote.Mail, s.Error.Message)
+			continue
+		}
+
+		mattermostUserID := usersByRemoteID[s.RemoteUserID].MattermostUserID
+		m.notifyUpcomingEvent(mattermostUserID, s.Events)
```
mickmister, line 126: 0/5 since we are sending a list of events. 

```suggestion
		m.notifyUserOfUpcomingEvents(mattermostUserID, s.Events)
```

server/mscalendar/availability.go
```diff
@@ -90,18 +90,56 @@ func (m *mscalendar) syncStatusUsers(userIndex store.UserIndex) (string, error)
...
+	}
+}
+
 func (m *mscalendar) setUserStatuses(users []*store.User, schedules []*remote.ViewCalendarResponse) (string, error) {
```
mickmister, line 130: I'm wondering if we want to stop using the variable name `schedules` since we are no longer using `GetSchedule` and this is not a `remote.ScheduleInfo` struct anymore. Maybe change to `responses`?

server/mscalendar/settings.go
```diff
@@ -35,6 +35,13 @@ func NewSettingsPanel(bot bot.Bot, panelStore settingspanel.PanelStore, settingS
 		store.UpdateStatusSettingID,
 		settingStore,
 	))
+	settings = append(settings, settingspanel.NewBoolSetting(
+		store.GetRemindersSettingID,
+		"Get Reminders",
+		"Do you want to get reminders for upcoming events?",
```
mickmister, line 41: 0/5. Then we would also change the setting to `ReceiveReminders`
```suggestion
		"Do you want to receive reminders for upcoming events?",
```

server/mscalendar/availability.go
larkox: Not in all places. We still need the ID to refer to the event when we talk with MSGraph. It should be used only on our side when we want to make sure it is the same event as other.

server/mscalendar/availability.go
larkox: 0/5 your proposal seems too verbose. What about just `notifyUpcomingEvents`?

server/mscalendar/availability.go
```diff
@@ -115,30 +115,30 @@ func (m *mscalendar) deliverReminders(users []*store.User, schedules []*remote.V
...
-func (m *mscalendar) setUserStatuses(users []*store.User, schedules []*remote.ViewCalendarResponse) (string, error) {
+func (m *mscalendar) setUserStatuses(users []*store.User, calendarViews []*remote.ViewCalendarResponse) (string, error) {
 	toUpdate := []*store.User{}
 	for _, u := range users {
 		if u.Settings.UpdateStatus {
 			toUpdate = append(toUpdate, u)
 		}
 	}
 	if len(toUpdate) == 0 {
-		return "No users want their status", nil
```
mickmister: idk about you, but I really just don't want my status 😛

server/mscalendar/views/calendar.go
```diff
@@ -20,22 +21,42 @@ func RenderCalendarView(events []*remote.Event, timeZone string) (string, error)
...
+}
+
+func renderEvent(event *remote.Event, asRow bool) (string, error) {
 	start := event.Start.Time().Format(time.Kitchen)
 	end := event.End.Time().Format(time.Kitchen)
 
-	return fmt.Sprintf("%s - %s `%s`", start, end, event.Subject)
+	format := "(%s - %s) [%s](%s)"
+	if asRow {
+		format = "| %s - %s | [%s](%s) |"
```
mickmister, line 51: Can you post a screenshot of these two formats? I'm still curious about rendering the single event as a "headerless" table.

server/plugin/plugin.go
```diff
@@ -218,6 +218,9 @@ func (p *Plugin) ServeHTTP(pc *plugin.Context, w http.ResponseWriter, req *http.
 
 func (p *Plugin) IsAuthorizedAdmin(mattermostUserID string) (bool, error) {
 	env := p.getEnv()
+	if mattermostUserID == env.BotUserID {
```
levb, line 221: This is weird, but makes sense

server/mscalendar/views/calendar.go
larkox: Table format:
Screenshot from 2020-04-24 15-48-04
Single event:
Screenshot from 2020-04-24 15-49-54

server/plugin/plugin.go
larkox: It is indeed weird. Merge error bringing back the changes to use the bot account. Will fix.

server/mscalendar/views/calendar.go
mickmister: I honestly like the table more. It might look even better without the headers. @asaadmahmood Any thoughts on this?

server/mscalendar/user.go
```diff
@@ -168,7 +169,18 @@ func (m *mscalendar) GetRemoteUser(mattermostUserID string) (*remote.User, error
 }
 
 func (m *mscalendar) IsAuthorizedAdmin(mattermostUserID string) (bool, error) {
-	return m.Dependencies.IsAuthorizedAdmin(mattermostUserID)
+	for _, userID := range strings.Split(m.AdminUserIDs, ",") {
```
mickmister, line 172: This is a pretty unexpected change, especially for the scope of the PR. Is there value lost by removing the injected dependency here? What packages had access to this functionality that no longer have access?

DHaussermann:
Tested this functionality including the following cases
- Decline status updates 
- Decline prompts on status updates  
-- Status updates to busy without prompt 
-- Status Return from busy without prompt 
- Update status without notifications when prompt is declined
- When meeting is canceled / moved after status already set to busy, the user is prompted to flip status back when sync notices they are no longer busy 
- No prompt on back to back or overlapping meeting as status does change from busy.

@larkox this looks really good! Though there are many possible edge cases (not necessarily related to status confirmation) that I will continue testing in master.

There were a couple small UX issues.
1. @mickmister noticed that when the status is changed after an event has already stated (invited to an in progress meeting for example)  the info in the prompt is misleading...
Notification-Event-in-progress
Technically accurate that there are no "upcoming" events but there is one in progress. The user was not able to be notified ahead of time. Can we add a a special case for this such that if the event is happening but stated in the past there could be text such as `You have an event ongoing. Shall I change your status...` ?  

2. Minor point - MS Calendar never refers to the event name as `subject` maybe drop off the `with subject` text from this dialogue? This way you have something like "Your event `Meet for Dinner` will start soon.
Event-Subject

asaadmahmood:
@larkox Based on your case, when a person changes his status to online, can we inform the plugin that the person manually made himself online, so that the next time it can ask me to change my status to dnd.

mickmister:
> the next time it can ask me to change my status to dnd.

@asaadmahmood So even if they have opted out of confirmations, we would still give them a chance to confirm? That is certainly feasible based on how we currently have things set up.

mickmister:
@DHaussermann Here's how I have it looking with some quick changes, including using a link for the event title


MS Calendar refers to the subject as a `title`, but in their data model it's called `subject`, so that's why we have it as that currently. Do you suggest we call it `title`?


Although the calendar also says `(No subject)` when an event has no title.


Here's how I have it looking when this is the case:


-----

Also to address your point of "when the status is changed after an event has already stated": Nothing should happen for the rest of that event. The user should not receive any prompts. It is supposed to take into account the fact that you changed your status during the meeting and assume you don't want your status changed at all until a new event comes along. I am unable to reproduce your situation.

-----

Another thing, when the event is currently ongoing, we still have access to the event and its subject. Should we have the event subject and link in the prompt, instead of your suggestion of the ambiguous `You have an event ongoing`?

mickmister:
@larkox Here's my code in case it is useful

```go
func renderScheduleItem(event *remote.Event, status string) string {
	if event == nil {
		return fmt.Sprintf("You have an event ongoing.\n Shall I change your status to %s?", prettyStatuses[status])
	}

	subject := "(No subject)"
	if event.Subject != "" {
		subject = event.Subject
	}

	link, _ := url.QueryUnescape(event.Weblink)
	fullSubject := fmt.Sprintf("`%s`", subject)
	if link != "" {
		fullSubject = fmt.Sprintf("[%s](%s)", subject, link)
	}

	resp := fmt.Sprintf("Your event %s will start soon.", fullSubject)
	resp += fmt.Sprintf("\nShall I change your status to %s?", prettyStatuses[status])
	return resp
}

```

asaadmahmood:
@mickmister We should not ask them again for events they have explicitly asked not to receive notifications for, but yes, it can be for any new events.

DHaussermann:
Thanks @mickmister Yes I'm in favor of both of these ideas. 
1. Yes, since the field is called subject and it does actually show `No subject` in the MS Calendar UI when empty, we could show the same in the notification.

2. If it's not a large amount of effort, we could show the name and link to the current on going event. I just wanted the problem solved where the user see's the unhelpful text of no upcoming events. That said, the UX would be odd if there was the name and the link because once the user selects and option would this info disappear and get replaced by `The status has not been changed.` or `The status has been changed to Do Not Disturb.` then the user can't click the link you just showed them.
We could maybe create a new UX issue to add a link to `viewcal` or `summary` in the post confirming your status was changed.  Maybe @asaadmahmood also has thoughts on this.

larkox:
@DHaussermann Let me state the expected behaviour, some edge cases and the reasoning behind it.

Case Update status automatically
- When a user is not DnD, and there is an upcoming event, the status will change to DnD.
- When the event ends and there are no ongoing or upcoming events (within 10 minutes if I remember correctly), it will change to online.
- When the event ends, if there are ongoing or upcoming events, the status will stay DnD until those end.
- If the user sets back the status to online during an event it will stay online until a new upcoming event appears.

Case Get confirmation for status update
- When a user is not DnD, and there is an upcoming event, it will be asked to change the status to DnD.
  - If the user says no or ignore the message, the user will not receive any new confirmation until a new event appears.
  - If the user says yes, the status will change to DnD.
- When the event ends and there is no other ongoing or upcoming event, the user will receive a message asking to change back to Online.
 - When the event ends, if there are ongoing or upcoming events, no status change confirmation until those end.
- If the user sets back the status to online during an event, it will not receive any new change confirmations until a new upcoming event appears.

Edge cases:
- User was DnD when the event started:
  - Will set back to online (or receive online confirmation) when the event ends. The plugin considers the user free after the event, even if the user was DnD before.

- User sets the status to online during the event, and then back to DnD during the event.
  - Will set back to online (or receive online confirmation) when the event ends. The plugin considers the user being DnD because of the event, and therefore consider it free when it finishes.

- Two meetings are apart by 5-10 minutes.
  - Will keep the status as DnD those 5-10 minutes. The plugin consider the new upcoming event to be in the same "busy frame".
  - If user sets the status to Online during those 5-10 minutes, the status will be kept as online along the duration of the second meeting. The plugin consider the new upcoming event to be in the same "busy frame", so it will consider the user set the online status for both.
  - This edge case is highly dependent on when the job runs. The job must run and find the two events, and then you have to change the status after then. For example:
    - Event A finishes at 13:55
    - Event B starts at 14:00
    - Job runs at 13:50. Since it looks for upcoming events 10 minutes ahead, it will find Event A and Event B.
    - Job will see that the user is DnD, so will consider Event B "taken care of"
    - User changes manually the status to online at 13:55
    - Job runs at 14:00. It only finds Event B, but since Event B has already been "taken care of", it does not do anything.

server/mscalendar/user.go
larkox: This was implemented on plugin.go, and passed to MSCalendar plugin, and only used from there. Therefore, implementing it here made more sense, and simplified the dependencies.

mickmister:
> If there is no subject, the name of the event is not "(No Subject)", it does not have a subject.

@larkox Makes sense :+1:

DHaussermann:
Thanks for detailing the cases @larkox This all seems to be working as expected. 

I did see once case where the leave the confirmation message had been displayed for hours and then selecting an option no longer worked. `Cannot check whether there is an event attached`
Confirmation1
I'm not sure how this occurred. Any thoughts?

I tested an edge case where the event is deleted before the recipient selects an option. and cases where the user has disconnected and reconnected their account before selecting an option. these both behave normally.

mickmister:
The `Cannot check whether there is an event attached` showing up is an error that "shouldn't ever happen" and can be taken care of in a separate ticket. It will require some investigation, but I don't think it's a blocker to get this merged. @DHaussermann What are your thoughts?

DHaussermann:
@mickmister Given that exploratory testing found no additional issues and I have made sure cases posted above by @larkox have all been covered now - I would be happy to merge this PR and investigate the remaining issue separately.

LGTM!

Issue #75, opened 2020-04-07
Remove dedicated bot Microsoft user


mickmister:
#### Summary

We are changing the `GetSchedule` API requests to use OAuth credentials from each individual user. This makes the benefits of the bot user nil because each request will be scoped to a specific user now. As a result, we are removing the dedicated bot user as it has no benefit and due to the complexity it brings to the system admin.

PR #76, opened 2020-04-08
Remove dedicated Microsoft integration user

mickmister:
#### Summary

This PR removes the bot account from the OAuth process. The application level permissions work the exact same way as before.

Calls to `GetSchedule` are now adjusted to one user's schedule, to have access to the user's event details.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/75


server/mscalendar/availability_test.go
```diff
@@ -92,25 +92,14 @@ func TestSyncStatusAll(t *testing.T) {
...
-
-			mockPluginAPI.EXPECT().GetMattermostUser("bot_mm_id").Return(&model.User{}, nil).Times(1)
-
-			mockRemote.EXPECT().MakeClient(context.Background(), token).Return(mockClient).Times(1)
-			mockClient.EXPECT().GetSuperuserToken().Return("bot_bearer_token", nil)
-			mockRemote.EXPECT().MakeSuperuserClient(context.Background(), "bot_bearer_token").Return(mockClient)
+			mockRemote.EXPECT().MakeSuperuserClient(context.Background()).Return(mockClient, nil)
 
-			mockClient.EXPECT().GetSchedule("bot_remote_id", []string{"user_email@example.com"}, gomock.Any(), gomock.Any(), 15).Return([]*remote.ScheduleInformation{tc.sched}, nil)
+			mockClient.EXPECT().GetSchedule(gomock.Any(), gomock.Any(), gomock.Any(), 15).DoAndReturn(
```
larkox, line 97: I am not a fan to use Any, but I understand this make the test simpler and not specially a worse test.

server/mscalendar/filters.go
```diff
@@ -41,16 +43,18 @@ func withClient(m *mscalendar) error {
...
 		return nil
 	}
-	client, err := m.MakeClient()
-	if err != nil {
-		return err
-	}
 
+	var client remote.Client
+	var err error
 	if m.actingUser.MattermostUserID == m.Config.BotUserID {
```
larkox, line 49: It is still necessary this BotUserID check here?

mickmister: Yes, if we are in a function that calls `withClient`, we want the bot user to be able to create its client without explicitly calling `withSuperuserClient`. This if statement takes care of that.

server/mscalendar/availability_test.go
mickmister: These arguments are structs, so they cannot be asserted inline. Instead, we call `DoAndReturn`, which allows us to examine the args that were passed in, and do our own assertions within the callback. I've added some more assertions here now.

DHaussermann:
@mickmister I started over with a fresh server and tested an install without a bot. 
Seems like everything works fine except 2 command... 

`/mscalendar viewcal`
`/mscalendar summary view`

They produce the following error: `msg":"Unable to execute command.","path":"/api/v4/commands/execute","request_id":"zzex7xgc8p8qdydunaunguto3e","ip_addr":"...","user_id":"g1a5x4iu9f883gmbwgemdxh6ch","method":"POST","err_where":"mscalendarplugin.ExecuteCommand","http_code":500,"err_details":"Command /mscalendar summary failed: 403 Forbidden: {\"error\":{\"code\":\"ErrorAccessDenied\",\"message\":\"Access is denied. Check credentials and try again.\"}}"}`

All other slash commands seem to work fine. 

If this is related to a bot-less install, possibly there is an update needed to how permissions are configured. Thoughts? 

Also Please remove  `Step 3: Configure Bot Account` from the read-me.

mickmister:
@DHaussermann I'm unable to reproduce the issues with `/mscalendar viewcal` and 
`/mscalendar summary view` on a fresh install. I've removed the bot account instructions from the README.

Issue #53, opened 2020-03-12
Accept/Decline a meeting invite in a chat window


larkox:
Whenever you receive an invite, the bot should send you a message with the meeting, and allow you to accept or decline the meeting from the message.

larkox:
@levb @mickmister After further inspection on the code, this is already done.

Right now it is using a Select in order to decide what to do with the event. On our previous talk, we talked about using buttons instead of the select, but here may be working nice.

The only improvement I would make on the code here is that once you have answered, the post is updated so you can no longer update the status, it will only state: "You have accepted this meeting". What do you think?

Also, keep in mind that this "You have accepted this meeting" would not be updated if the user updates the status in the calendar.

levb:
@larkox all i can say is, "let's get it to maximum accordance with the spec". I initially implemented it with buttons, but then switched to a select to be able to reflect the current acceptance status.

PR #79, opened 2020-04-08
[GH-53] Accept or decline meetings from the notification message

larkox:
#### Summary
Improve the behaviour for accepting or declining meetings, and improve the behaviour of notifications.

#### Ticket Link
#53


server/api/post_action.go
```diff
@@ -5,35 +5,36 @@ package api
...
+	"strings"
 
 	"github.com/mattermost/mattermost-server/v5/model"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/config"
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/mscalendar"
 )
 
-func (api *api) preprocessAction(w http.ResponseWriter, req *http.Request) (mscalendar.MSCalendar, *mscalendar.User, string, string) {
+func (api *api) preprocessAction(w http.ResponseWriter, req *http.Request) (mscalendar.MSCalendar, *mscalendar.User, string, string, string) {
```
mickmister, line 16: Can we name these return values, so it is obvious what each value is meant for?

server/api/post_action.go
```diff
@@ -70,14 +71,47 @@ func (api *api) postActionTentative(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	sas := p.Attachments()
+	if len(sas) == 0 {
+		http.Error(w, "Failed to update the post: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	sa := sas[0]
+	sa.Actions = mscalendar.GetPostActionSelect(eventID, response, req.URL.String())
```
mickmister, line 109: It doesn't seem obvious to me that `GetPostActionSelect` is specific to responding to an event. Maybe rename to `NewPostActionForEventResponse`

server/mscalendar/notification.go
```diff
@@ -248,34 +259,70 @@ func (processor *notificationProcessor) updatedEventSlackAttachment(n *remote.No
 		return false, nil
 	}
 
+	allChanges := append([]string{}, added...)
+	allChanges = append(allChanges, updated...)
+	allChanges = append(allChanges, deleted...)
```
mickmister, line 264: Just a suggestion, up to you if you want to make a change here. It's an extra line but I prefer the first one here:

```suggestion
	var allChanges []string
	allChanges = append(allChanges, added...)
	allChanges = append(allChanges, updated...)
	allChanges = append(allChanges, deleted...)
```

or

```suggestion
	allChanges := append(added, updated...)
	allChanges = append(allChanges, deleted...)
```

DHaussermann:
@larkox 
Unless there are extra steps I'm not aware of - I'm not seeing this work. 
Possibly this is not testable without https://github.com/mattermost/mattermost-plugin-mscalendar/pull/44 ? As part of that PR I can enable the bot delivering me invite notifications. 
Thoughts?

larkox:
@DHaussermann in order to test this:

1. (In Mattermost) Log in to MSCalendar with account A
2. (In Mattermost) Add a subscription (/mscalendar subscribe)
3. (In Outlook) Create a new event with a account B, and invite the account A
4. (In Mattermost) You should receive a message with the event, and you can answer to the invite.
5. (In Outlook) Change any relevant information (time, subject or location)
6. (In Mattermost) You should receive a message with the relevant change
7. (In Outlook) Change non relevant information (attendees, attendees status, ...)
8. (In Mattermost) You should not receive any new message
9. (In Outlook) Create a new event with account A
10. (In Mattermost) You should receive a new message but without the actions (you cannot answer to your own events)

DHaussermann:
@larkox I don't see the expected for `(In Mattermost) You should receive a message with the event, and you can answer to the invite.` As user A
- I see the ephemeral post confirming a subscription was created 
- I see the event invite is MS Calendar that I was invited to 
- I do not see any message in Mattermost about the event.  (it is visible with `viewcal`)

I redeployed after the merge in from Master earlier. Please advise on further isolation steps.

larkox:
@DHaussermann It feels weird. It is working for me, but just after a while after creating the subscription. It feels like sometimes MS just do not send the notification (no apparent reason). @mickmister have you encountered this before? Any idea about this?

levb:
It has happened to me several times before; but if it's happening now @DHaussermann should be able to retry; or we can do a zoom with @larkox or me and debug online.

larkox:
Another side note, the subscriptions right now have a short lifetime (made bigger in a different PR if I am not mistaken). Make sure you have an active subscription by typing `/mscalendar subscribe list`

DHaussermann:
@larkox Your suspicion above was correct that deliveries will not start right away. Also seem like if I wait over 15 min the subscription is gone so it was hard to see this work.  (I am aware of Lev's  PR to extend and renew them) 

That being said maybe we could still connect on this. I can get the the event message when it is first sent but for a 1st pass - I see some other issues.

1. Updates made by the sender did not deliver a message for a relevant detail like time change
2. Accepting using the the drop down in the event post did not update the acceptance status on the MS Calendar side.
3. All data about the event is enclosed with `[]` which are rendering in the post. 
EventMsg

larkox:
@DHaussermann I tracked down several of the problems.

> Updates made by the sender did not deliver a message for a relevant detail like time change

I found out that the delays also happen with the updates. You must see a call in ngrok (or any other traffic log of the server) to `/plugins/com.mattermost.mscalendar/notification/v1/event`. If you do not see this, is because Microsoft has not send the notification yet.

> Accepting using the the drop down in the event post did not update the acceptance status on the MS Calendar side.

The first one should be accepted. Subsequent changes are not accepted due to Microsoft changing the id of the event. The only solution for this would imply several changes and bad user experience in my opinion, so I will make so once you have decided something, you cannot change it in Mattermost.

> All data about the event is enclosed with [] which are rendering in the post.

The formatting of these messages is not on the scope of this PR.

DHaussermann:
@larkox I'm afraid I don't have much info to add about these 2 issues. 

1. Regarding no message about event updates, I don't see any activity like what you described on my server when I'm waiting for the DM.

2. Regarding the attendance status, I am selecting the affirmative option of `Yes` in the drop down and I still see `RSVP` as the state in Outlook. I have waited sevral minutes for the update to take place. 

I am also noticing another issue which may be the same as what I'm seeing in 1. For some reason, only the 1st event invite is delivering an event to me in a DM. In the past I saw this and assumed it was caused by the  subscription time to live. But, now I think I'm seeing an issue which may also explain why I'm not getting event updated DMs. Can you try and repo this...
Steps:
 - As _User A_ Create a subscription
- Wait 5 min to ensure the subscription is working (this was discussed above - not sure if I need to create a separate issue for this delay)
- As _User B_ invite _User A_ to an event from outlook
- Wait a few seconds and note that _User A_ has received a DM with event details
- As _User B_ invite _User A_ to another event from outlook 
- Use `/mscalendar subscribe list` to confirm subscription is still in place. 
Expected: 2nd event delivered in a DM 
Observed: No DM received after 2nd invite 

Please advise on next steps for this. Maybe we could set a time to connect if we aren't seeing the same results.
Also curious about your thoughts on if any of these issues seem as thought they may be related to bugs already fixed in master.

levb:
@DHaussermann Sorry for a superficial comment in a hurry. I think only some of the event updates trigger subscription messages in MSGraph. I am 100% sure that changing the time of the event does, but I am pretty sure that editing the details does not. Have not looked for the detailed documentation on what changes are broadcast.

DHaussermann:
@larkox just a heads-up I'll need to redeploy and test this again. I'll let you know if I see any change.

larkox:
@DHaussermann I found the error with the reply to the events. It should be fixed now. It might be related to the other error of new events not showing, but I am not really sure about that. It feels like notifications come from Microsoft at different speeds (I made one event, received nothing, I made a second event, received the second, and shortly after I received the first). There is not much we can do on that front, apparently. Nevertheless, it is working on my end, so whenever you are going to give it another run, ping me and I can see if there is anything different in your environment that might make this happen.

DHaussermann:
@larkox I have tested this again.
I now see the behavior where accept the ephemeral post now updates and the status is reflected in MS Calendar. 🎉 

Yes I am also seeing that the events DMs are no delivered reliably. If I create several invites in a row, they are not delivered at the same time or chronologically.  

I still have not seen a DM for an event update though.

DHaussermann:
@larkox  quick follow-up I do see updates working! They were all delivered at once about 10 minutes later. 

Last question on this PR - Selecting anything other than `Yes` for attendance satus does not work. I believe this is out of scope. correct?

Issue #54, opened 2020-03-13
Renew event subscriptions periodically


mickmister:
#### Summary

Microsoft Graph's subscriptions have a limited lifespan, with a maximum of 2 days that we can assign. I'm thinking renewing every 24 hours, and checking if we need to renew upon start-up, is a reasonable plan. We can renew 20 subscriptions with each http request.

User intervention is not required to create or renew a subscription, so it's not detrimental if a subscription expires, due to e.g. a MM server being down for 2 days for whatever reason.

mickmister:
@levb Mind if I take this one, unless you've already begun work on it?

PR #82, opened 2020-04-11
Added a job to renew user subscriptions

levb:
#### Summary
- Added a simple job to renew all user subscriptions every 24 hours
- Increased the default subscription expiry to 48 hours

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/54

### DIFF
--- a/server/jobs/job_manager.go
+++ b/server/jobs/job_manager.go
@@ -70,7 +70,7 @@ func (jm *JobManager) OnConfigurationChange(env mscalendar.Env) error {
 
 	jm.registeredJobs.Range(func(k interface{}, v interface{}) bool {
 		job := v.(RegisteredJob)
-		enabled := job.isEnabledByConfig(env)
+		enabled := (job.isEnabledByConfig == nil) || job.isEnabledByConfig(env)
 		_, active := jm.activeJobs.Load(job.id)
 
 		// Config is set to enable. Job is inactive, so activate the job.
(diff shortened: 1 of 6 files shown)


server/jobs/renew_job.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewRenewJob() RegisteredJob {
+	return RegisteredJob{
+		id:       "renew",
+		interval: 24 * time.Hour,
+		work:     runRenewJob,
+	}
+}
+
+// runDailySummaryJob delivers the daily calendar summary to all users who have their settings configured to receive it now
```
mickmister, line 22:
```suggestion
// runRenewJob calls renews the event subscription for each connected user
```

server/jobs/job_manager.go
```diff
@@ -70,7 +70,7 @@ func (jm *JobManager) OnConfigurationChange(env mscalendar.Env) error {
 
 	jm.registeredJobs.Range(func(k interface{}, v interface{}) bool {
 		job := v.(RegisteredJob)
-		enabled := job.isEnabledByConfig(env)
+		enabled := (job.isEnabledByConfig == nil) || job.isEnabledByConfig(env)
```
larkox, line 73: Why not this `or` logic inside the "isEnableByConfig" logic?

levb: 'cause providing a function for the "always on" case seemed wrong. nil is more appropriate?

Issue #71, opened 2020-04-02
Delete OAuth state from KV store after OAuth process is complete


mickmister:
#### Summary

Currently the OAuth state data is left persisting in the KV store after the when it should be removed when the authentication process is finished.

avddvd:
@mickmister I would like to work on this issue. 

I thought of couple of ways to manage OAuth2 state

1. Use `StoreTTL(key string, data []byte, ttlSeconds int64)` instead of `Store(key string, data []byte)` with an appropriate time to live value.

2. Call `Delete(key string)` right after state is verified with `VerifyOAuth2State`

@mickmister I would think approach 1 is cleaner once an appropriate time to live value is agreed on.

Please suggest

mickmister:
Hi @avddvd, thanks for picking up this issue!

Using option one would ensure it gets removed regardless of what happens after storing, so that sounds good to me. The OAuth process is virtually instantaneous, barring user intervention time. I think 5 minutes is an appropriate, generous time window.

mickmister:
@avddvd To add this, I suggest responding with a user-friendly message when they missed the time window. Something like "Your authentication token has expired. Please try and reconnect."

avddvd:
@mickmister When user misses the time window (i.e: takes more than 5 minutes to finish OAuth), the following happens 
- managed OAuth state gets deleted
- as a result, `VerifyOAuth2State` returns an error specifically `AppError`

As far as my understanding of OAuth flow, the above scenario forces user to reconnect (restart OAuth)

Would you suggest to return `errors.New("Your authentication token has expired. Please try and reconnect.")` message from `VerifyOAuth2State` ?

mickmister:
> Would you suggest to return `errors.New("Your authentication token has expired. Please try and reconnect.")` message from `VerifyOAuth2State`?

Yeah that sounds good to me :+1:

mickmister:
@avddvd ^

avddvd:
Hey @mickmister 

This check is already in place
`	if string(data) != state {
		return errors.New("authentication attempt expired, please try again")
	}`

PR #83, opened 2020-04-14
use StoreTTL in place of Store for managing OAuth2 state

avddvd:
#### Summary
- Use StoreTTL method of KVStore to manage the OAuth2 state.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/71


larkox:
@mickmister The state of the OAuth has to live from the moment the user hits the link, until the login process is finished. Therefore, you have to take in account the time the user takes to put their credentials. I think 5 minutes is a reasonable time.

server/store/oauth2_store.go
```diff
@@ -18,11 +23,11 @@ func (s *pluginStore) VerifyOAuth2State(state string) error {
 		return err
```
mickmister, line 23: @avddvd Moving the conversation on the issue https://github.com/mattermost/mattermost-plugin-mscalendar/issues/71#issuecomment-615768566 to here since the code is here.

I could be entirely wrong but I think `string(data) != state` would occur if someone is impersonating Microsoft or the user tries to authenticate twice or something. That error can stay the way that you currently have it :+1:

If `err == store.ErrNotFound` in the block below, then we know specifically that the state is not in store, and can assume this means the Expiry has occurred. So, my only suggestion is that we add this block:

```suggestion
		if err == store.ErrNotFound {
			return errors.new("authentication attempt expired, please try again")
		}
		return err
```

avddvd: Thanks @mickmister for the clarification, pushed a new change to check error in case of key not found (i.e: since TTL expired )

mickmister: This `err == store.ErrNotFound` check needs to be in the `err != nil` block as my suggestion above has it at `line 25`. At the point of `line 28` where you currently have the check, the function would have already returned due to what's currently at `line 25`.

Also, I thought about it, and the `string(data) != state` error should be something more explicit like `invalid oauth state`.

avddvd: Fixed this, oversight from my end

mickmister: Sorry for being so persistent about this (it's especially a pain because it's a small PR), but what do you think about changing the error string for the `string(data) != state` check to be `invalid oauth state`?

avddvd: No worries, Done

mickmister:
Thanks @avddvd!

mickmister:
@avddvd Make sure to join the [community server](https://community.mattermost.com) to work more closely with the team! We have a [channel](https://community.mattermost.com/core/channels/plugin-microsoft-calendar) for this project on the server.

PR #84, opened 2020-04-14
POC for plugins KVStore cache

larkox:
#### Summary
KVStore does not have at the moment any cache system on it, as it understands the plugins are the ones taking in consideration what needs to be cached and what not. Since the latest discussions on "individual records vs. array of records" on the kv store, I thought this might be an important path to consider.

Therefore, I have created a proof of concept for a cache layer on the kvstore implemented in utils. The usage would be simple: any store that we already have created but want it to work with caching, we just have to wrap it around the cache layer:
```
kvstore.NewCachedStore(kvstore.NewHashedKeyStore(basicKV, UserKeyPrefix), "userCache", 5000),
```

This implementation still has many flaws and has been almost not tested, but I think it is a good starting point for discussion.

#### Ticket Link
None

### DIFF
--- a/server/store/store.go
+++ b/server/store/store.go
@@ -50,7 +50,7 @@ func NewPluginStore(api plugin.API, logger bot.Logger) Store {
 	basicKV := kvstore.NewPluginStore(api)
 	return &pluginStore{
 		basicKV:            basicKV,
-		userKV:             kvstore.NewHashedKeyStore(basicKV, UserKeyPrefix),
+		userKV:             kvstore.NewCachedStore(kvstore.NewHashedKeyStore(basicKV, UserKeyPrefix), "userCache", 5000),
 		userIndexKV:        kvstore.NewHashedKeyStore(basicKV, UserIndexKeyPrefix),
 		mattermostUserIDKV: kvstore.NewHashedKeyStore(basicKV, MattermostUserIDKeyPrefix),
 		subscriptionKV:     kvstore.NewHashedKeyStore(basicKV, SubscriptionKeyPrefix),
--- a/server/utils/kvstore/cache_layer.go
+++ b/server/utils/kvstore/cache_layer.go
@@ -1,0 +1,203 @@
+package kvstore
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/mattermost/mattermost-server/v5/model"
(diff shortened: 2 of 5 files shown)


levb:
A few general thoughts, not specific to the implementation here.
1. 0/5 I am sort of against trying to solve caching across multiple operations, especially considering the HA implications, at this point. Open to discuss it.
2. In solar lottery, I already added a cache/memory-based store which is useful for intra-request performance improvements and testing. https://github.com/mattermost/mattermost-plugin-solar-lottery/blob/master/server/utils/kvstore/cache.go - would be great to reconcile. 
3. 1/5 I think that intra-request it makes sense to cache go objects (instead of bytes?) and avoid repetitive RPC trips to get things like user records, in addition to KV.

larkox:
@levb 

> 0/5 I am sort of against trying to solve caching across multiple operations, especially considering the HA implications, at this point. Open to discuss it.

This approach allow the implementation to decide which "stores" are cached and which not. This way, you can have the parts you mostly read cached, and more changing parts not cached. Nevertheless, HA implications are there and at the moment are really hard to solve. "Fast" communication between clusters for plugins (to do all the cache invalidation) would be a desirable, but I am not sure whether we can hope or dream for that.

> In solar lottery, I already added a cache/memory-based store which is useful for intra-request performance improvements and testing. https://github.com/mattermost/mattermost-plugin-solar-lottery/blob/master/server/utils/kvstore/cache.go - would be great to reconcile.

The mattermost server already has some cache code that can be reused, including a LRU implementation with expires, if I am not mistaken. Maybe it is easier to use that one (0/5).
Nevertheless, the kind of cache you are proposing here does not work for a general case due to HA implications, and due to that it does not work specially well for the mscalendar situation we have at hand (neither does work perfectly the solution I am proposing).

> 1/5 I think that intra-request it makes sense to cache go objects (instead of bytes?) and avoid repetitive RPC trips to get things like user records, in addition to KV.

You are right, as long as we do not want to check equality. It is easier to check equality over []byte than over go objects.

larkox:
Having a cache in plugins without being able to invalidate cache through the cluster seems not efficient, and any kind of cache should be made in a case by case basis. Therefore we can dismiss this PR.

PR #85, opened 2020-04-15
Fix logging calls

mickmister:
#### Summary

It turns out all of the issues were introduced by me :slightly_smiling_face:. For some reason I thought we could use the syntax `Debugf("An error occurred", "error", err.Error())`

#### Ticket Link

#69


server/jobs/job_manager.go
```diff
@@ -77,15 +77,15 @@ func (jm *JobManager) OnConfigurationChange(env mscalendar.Env) error {
 		if enabled && !active {
 			err := jm.activateJob(job)
 			if err != nil {
-				jm.env.Logger.Errorf("Error activating job", "id", job.id, "error", err.Error())
+				jm.env.Logger.Errorf("Error activating %s job. %s", job.id, err.Error())
```
levb, line 80: total nit, but why not `%v` and `err`? Just a little shorter...

levb:
@mickmister write your own for that :) (pairs)

server/jobs/job_manager.go
mickmister: @levb Fixed, and did a quick traversal of the codebase to fix others

mickmister:
> @mickmister write your own for that :) (pairs)

@levb Not sure if this is what you meant but I've created a HW ticket https://github.com/mattermost/mattermost-plugin-mscalendar/issues/86

server/jobs/job_manager.go
```diff
@@ -77,15 +77,15 @@ func (jm *JobManager) OnConfigurationChange(env mscalendar.Env) error {
 		if enabled && !active {
 			err := jm.activateJob(job)
 			if err != nil {
-				jm.env.Logger.Errorf("Error activating job", "id", job.id, "error", err.Error())
+				jm.env.Logger.Errorf("Error activating %s job. %v", job.id, err)
```
larkox, line 80: Not sure if there is a stronger reasoning for this here, but in some messages you finish the sentence with a dot, and then the error, in some messages you finish the sentence with a colon, and then the error. (0/5) Wouldn't it be better to have them all following the same pattern?

Issue #100, opened 2020-04-27
Improve rendering of event notifications


larkox:
When a new event is created, the plugin shows a notification.

This notification at the moment shows an unfriendly interface. This task is to improve the look of this notifications.

PR #102, opened 2020-04-27
Improve notification looks

larkox:
#### Summary
Improve the notifications look.

Current status:
Screenshot from 2020-04-27 13-49-39
Screenshot from 2020-04-27 13-52-00
Screenshot from 2020-04-27 13-56-17

#### Ticket Link
#100


mickmister:
Excellent work @larkox! Do you think it makes sense to put the end time in the notification as well, or would it be unnecessary noise? Personally I would want to know the end time, on inspection of the notification.

larkox:
@asaadmahmood @aaronrothschild Any thoughts on @mickmister 's comment?

asaadmahmood:
@mickmister @larkox Yes, I think we should put the endtime, that's what a lot of calendar apps do.
We should also add something about the event repeating (if it is repeating). 
Screenshot 2020-04-28 at 1 01 36 PM

levb:
@larkox not merging yet if you want to fix the suggestions, or merge yourself as is?

larkox:
This is how currently looks:
Screenshot from 2020-04-29 15-20-13 - 1
Adding whether is a repeating event or not seems not to be as straightforward, so I think should be solved in a new ticket. I will create the issue.

asaadmahmood:
Not even sure if we need the duration now, I think we can remove it.

mickmister:
I agree, I think we should remove the duration in favor of the end time.

PR #118, opened 2020-05-06
[GH-89] Allow status change to support DND and Away

### LINKED ISSUE #89: Allow status change to support DND and Away
Some users want to be marked as busy when in a meeting, but also want to continue receiving Mattermost notifications. Having the choice be a configurable value for each user would allow us to do that. DND blocks all notifications. The user can configure their Mattermost account settings to allow Away to allow or disallow notifications, so in this case the user would want to allow notifications.

larkox:
#### Summary
Allow status change to support DND and Away. Settings added to both settings and login flow.

#### Ticket Link
Fix #89

### DIFF
--- a/server/mscalendar/availability.go
+++ b/server/mscalendar/availability.go
@@ -181,6 +181,10 @@ func (m *mscalendar) setUserStatuses(users []*store.User, calendarViews []*remot
 
 func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus string, res *remote.ViewCalendarResponse) (string, error) {
 	events := filterBusyEvents(res.Events)
+	busyStatus := model.STATUS_DND
+	if user.Settings.ReceiveNotificationsWhileOnMeeting {
+		busyStatus = model.STATUS_AWAY
+	}
 
 	if len(user.ActiveEvents) == 0 && len(events) == 0 {
 		return "No events in local or remote. No status change.", nil
@@ -188,14 +192,14 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
 
 	var message string
 	if len(user.ActiveEvents) > 0 && len(events) == 0 {
-		if currentStatus == model.STATUS_DND {
+		if currentStatus == busyStatus {
 			err := m.setStatusOrAskUser(user, model.STATUS_ONLINE, events)
 			if err != nil {
 				return "", err
 			}
 			message = "User is no longer busy in calendar. Set status to online."
 		} else {
-			message = "User is no longer busy in calendar, but is not set to DND. No status change."
+			message = "User is no longer busy in calendar, but is not set to busy. No status change."
 		}
 		err := m.Store.StoreUserActiveEvents(user.MattermostUserID, []string{})
 		if err != nil {
@@ -215,14 +219,14 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
 
 	if len(user.ActiveEvents) == 0 {
 		var err error
-		if currentStatus == model.STATUS_DND {
+		if currentStatus == busyStatus {
 			err = m.Store.StoreUserActiveEvents(user.MattermostUserID, remoteHashes)
 			if err != nil {
 				return "", err
 			}
 			return "User was already marked as busy. No status change.", nil
 		}
-		err = m.setStatusOrAskUser(user, model.STATUS_DND, events)
+		err = m.setStatusOrAskUser(user, busyStatus, events)
 		if err != nil {
 			return "", err
 		}
@@ -230,7 +234,7 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
 		if err != nil {
 			return "", err
 		}
-		return "User was free, but is now busy. Set status to DND.", nil
+		return "User was free, but is now busy. Set status to busy.", nil
 	}
 
 	newEventExists := false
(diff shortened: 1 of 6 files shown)


server/mscalendar/availability.go
```diff
@@ -215,22 +219,22 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
 
 	if len(user.ActiveEvents) == 0 {
 		var err error
-		if currentStatus == model.STATUS_DND {
+		if currentStatus == busyStatus {
```
mickmister, line 222: Hmm if I have my busy preference set to `Away`, and I'm currently set to `DND`, it will change me to `Away`. I'm not entirely sure this makes sense, because I'm already set to the busiest of busy.

@asaadmahmood For context, the user needs to be able to choose if they want to receive notifications while being marked "In a meeting". We are doing this by letting them choose between `DND` or `Away` in their plugin preferences.

server/mscalendar/availability.go
```diff
@@ -215,22 +219,22 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
...
+		err = m.setStatusOrAskUser(user, busyStatus, events)
 		if err != nil {
 			return "", err
 		}
 		err = m.Store.StoreUserActiveEvents(user.MattermostUserID, remoteHashes)
 		if err != nil {
 			return "", err
 		}
-		return "User was free, but is now busy. Set status to DND.", nil
+		return "User was free, but is now busy. Set status to busy.", nil
```
mickmister, line 237: Can you inject the `busyStatus` variable here? As well as the rest of the strings mentioning `set status to busy`. Also keep in mind these are debug strings, and only show up when using `/availability` as opposed to the scheduled status sync job.

server/mscalendar/settings.go
```diff
@@ -38,6 +38,13 @@ func NewSettingsPanel(bot bot.Bot, panelStore settingspanel.PanelStore, settingS
 		store.UpdateStatusSettingID,
 		settingStore,
 	))
+	settings = append(settings, settingspanel.NewBoolSetting(
+		store.ReceiveNotificationsWhileOnMeeting,
+		"Receive notifications while on meetings",
+		"Do you want to still receive Mattermost notifications while you are on a meeting?",
```
mickmister, line 44: I think they should be told explicitly what status they will be set to when they are asked this question. "If you select yes, we will set your status to `Away`, otherwise we will set it to `Do Not Disturb`.

server/mscalendar/welcome_flow.go
```diff
@@ -68,6 +68,14 @@ func (wf *welcomeFlow) makeSteps() {
 		FalseButtonMessage:   "No - Update my status automatically",
 		TrueResponseMessage:  "Cool, we'll also send you confirmations before updating your status.",
 		FalseResponseMessage: "Cool, we'll update your status automatically with no confirmation.",
+	}, &flow.SimpleStep{
+		Title:                "Receive notifications while on meetings",
+		Message:              "Do you want to receive notifications while you are on a meeting?",
+		PropertyName:         store.ReceiveNotificationsWhileOnMeetingName,
+		TrueButtonMessage:    "Yes - I would like to set my status to Away during meetings to continue receiving notifications.",
+		FalseButtonMessage:   "No - I would like to set my status to Do Not Disturb during meetings to not receive notifications.",
```
mickmister, line 76: I see we do explain it here though. I think the wording sounds good :+1:

server/store/flow_store.go
```diff
@@ -3,9 +3,10 @@ package store
 import "fmt"
 
 const (
-	UpdateStatusPropertyName    = "update_status"
-	GetConfirmationPropertyName = "get_confirmation"
-	SubscribePropertyName       = "subscribe"
+	UpdateStatusPropertyName               = "update_status"
+	GetConfirmationPropertyName            = "get_confirmation"
+	ReceiveNotificationsWhileOnMeetingName = "receive_notifications"
```
mickmister, line 8: Not sure if we can make the `value` shorter while still being clear. Any changes should apply to other similarly named vars of course

```suggestion
	ReceiveNotificationsDuringMeetingName = "receive_notifications_during_meeting"
```

hanzei, line 8: Nit: By `type`ing these it would be more clear that these form a group.

server/mscalendar/availability.go
larkox: As discussed through zoom, if the user want to be away during meetings, it will change to away.

server/store/flow_store.go
larkox: I added your suggestion. If you have any other idea to make it shorter but giving the same information, please let me know and I will change it.

larkox: Not sure what you mean with this. Do you mind to elaborate?

mickmister: I'm think he means something like

```go
type SettingProperty string

const (
	UpdateStatusPropertyName SettingProperty = "update_status"
```

mickmister: Then anywhere else in the code, you can reference these values as `SettingProperty` instead of `string`

hanzei: Exactly that @mickmister. Thanks for making this clear.

mickmister: You've changed the name of the var, but the main thing I was suggesting was to change the value to `receive_notifications_during_meeting`. The value `receive_notifications` is too vague imo and I would assume by default that it is referring to receiving event notifications in general.

server/mscalendar/availability.go
```diff
@@ -199,7 +199,7 @@ func (m *mscalendar) setStatusFromCalendarView(user *store.User, currentStatus s
 			}
 			message = "User is no longer busy in calendar. Set status to online."
 		} else {
-			message = "User is no longer busy in calendar, but is not set to busy. No status change."
+			message = fmt.Sprintf("User is no longer busy in calendar, but is not set to busy(%s). No status change.", busyStatus)
```
mickmister, line 202: Non-blocking nit, I'm thinking there should be a space between "busy" and the busyStatus

```suggestion
			message = fmt.Sprintf("User is no longer busy in calendar, but is not set to busy (%s). No status change.", busyStatus)
```

server/store/flow_store.go
larkox: Making this change now would imply touching several files (6 + mocks), and feel a bit out of this PR scope. I will take note of this improvement, and apply it to the extracted flow and settings.

PR #120, opened 2020-05-06
[GH-103] Add event reminder subscriptions to settings panel

### LINKED ISSUE #103: Add event reminder subscriptions to settings panel
We should add "Receive Meeting Invites" into the settings pane so users can change this setting after the login workflow.

This would be friendlier than having to use subscribe and makes sense sine all other decisions in login flow can be changed via the panel already. It feels slightly confusing that this option is missing when you open the panel after going through the login flow.

larkox:
#### Summary
Add event reminder subscriptions to settings panel

#### Ticket Link
Fix #103


server/mscalendar/settings_notifications.go
```diff
@@ -0,0 +1,137 @@
...
+func NewNotificationsSetting(getCal func(string) MSCalendar) settingspanel.Setting {
+	os := &notificationSetting{
+		title:       "Receive notifications of new events",
+		description: "Do you want to subscribe to new events and receive a message when they are created?",
+		id:          "notification_setting",
+		dependsOn:   "",
+		getCal:      getCal,
+	}
+
+	return os
```
mickmister, line 27: Why not just return without assigning?

server/mscalendar/settings_notifications.go
```diff
@@ -0,0 +1,137 @@
...
+	getCal      func(string) MSCalendar
+}
+
+func NewNotificationsSetting(getCal func(string) MSCalendar) settingspanel.Setting {
+	os := &notificationSetting{
+		title:       "Receive notifications of new events",
+		description: "Do you want to subscribe to new events and receive a message when they are created?",
+		id:          "notification_setting",
+		dependsOn:   "",
+		getCal:      getCal,
```
mickmister, line 24: We call this `getCalendar` in the function in `settings.go`. `getCal` sounds good to me to use in both places.

server/mscalendar/settings_notifications.go
```diff
@@ -0,0 +1,137 @@
...
+	id          string
+	dependsOn   string
+	getCal      func(string) MSCalendar
+}
+
+func NewNotificationsSetting(getCal func(string) MSCalendar) settingspanel.Setting {
+	os := &notificationSetting{
+		title:       "Receive notifications of new events",
+		description: "Do you want to subscribe to new events and receive a message when they are created?",
+		id:          "notification_setting",
```
mickmister, line 22: `notification_setting` is pretty vague. We use the word "notification" for pretty much any DM sent to the user by a bot. What about `new_or_updated_event_setting`?

server/mscalendar/settings_notifications.go
```diff
@@ -0,0 +1,137 @@
...
+	_, err := cal.LoadMyEventSubscription()
+	if err == nil {
+		return cal.DeleteMyEventSubscription()
+	}
+	return nil
+}
+
+func (s *notificationSetting) Get(userID string) (interface{}, error) {
+	cal := s.getCal(userID)
+	_, err := cal.LoadMyEventSubscription()
```
mickmister, line 59: So we aren't storing their choice in the settings store, and instead are relying on the stored subscription. I suppose this is good so we only have one source of truth and aren't constantly checking if both values are set (setting and subscriptionID) :+1:

server/mscalendar/settings_notifications.go
```diff
@@ -0,0 +1,137 @@
...
+		Title:   title,
+		Text:    text,
+		Actions: actions,
+	}
+
+	return &sa, nil
+}
+
+func (s *notificationSetting) IsDisabled(foreignValue interface{}) bool {
+	return foreignValue == "false"
```
mickmister, line 136: Maybe reference these as constants like `BOOL_STRING_FALSE`?

levb:
(not removing the Dev Review yet for @mickmister's feedback)

server/mscalendar/settings_notifications.go
larkox: I am making in parallel some changes on how to handle slack attachment context, and therefore how to pass this value here. Long story short, marshaling and unmarshaling the values as jsons. You can see some of that implementation [here](https://github.com/larkox/mattermost-plugin-utils/blob/FreeText/flow/steps/simple_step.go).

I will leave this as it is, and change it in the future when those changes are in place.

DHaussermann:
Tested and passed
- Confirmed that when no subscription is preset `Yes` will create one
- Selecting `No` from the panel will delete the subscription
- Briefly regression tested login work flow and working subscription
LGTM!

@larkox note that I did see an issue here but, it was not related to the settings panel.  Created https://github.com/mattermost/mattermost-plugin-mscalendar/issues/148 to address.

PR #121, opened 2020-05-06
Ensure subject is shown as "(No subject)" where appropriate

### LINKED ISSUE #109: Ensure all placed where Subject is used, that a blank Subject has decent UX
One situation in particular seemed off in the Daily Summary view. The table cell for subject is blank, so there is no clickable link to the event. suggest putting `(No subject)` there instead.

mickmister:
#### Summary

The two places I didn't convert are:
- The sentences where we say "Without a subject"
- When we create the `PostAction.Integration.Context`, and use the subject in the URL

#### Ticket Link

Fixes #109

### DIFF
--- a/server/mscalendar/notification.go
+++ b/server/mscalendar/notification.go
@@ -14,6 +14,7 @@ import (
 	"github.com/mattermost/mattermost-server/v5/model"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/config"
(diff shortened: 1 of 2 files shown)


DHaussermann:
I tried testing this but the plugin thinks no users need the summary delivered. 
Will test once https://github.com/mattermost/mattermost-plugin-mscalendar/pull/133 is fixed and merged as this is the issue blocking testing.

PR #122, opened 2020-05-06
Remove references to "common"

### LINKED ISSUE #114: Remove mentions of `common` deployment that does not use Azure AD
That Admin UI has references to a `common` deployment type that do not use Azure AD. These should be removed from the UI.
Common
As far as I know, the only other place this needs to be removed is in the read-me.

mickmister:
#### Summary

A made a PR for this a while ago but missed the `plugin.json` file.

#### Ticket Link

Fixes #114

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -61,10 +61,9 @@
             },
             {
                 "key": "OAuth2Authority",
-                "display_name": "Azure Directory (tenant) ID, or \"common\"",
+                "display_name": "Azure Directory (tenant) ID",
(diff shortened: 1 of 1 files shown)


levb:
Just for my understanding, do we still support personal Outlook.com accounts? Just removing the confusing references? Or is it no longer supported at all?

mickmister:
@levb I believe it is not supported now, though I haven't once tried to set it up that way. It this a a use case we want to support?

PR #124, opened 2020-05-07
Use external flow, bot and settings panel

larkox:
#### Summary
Remove all flow, panel settings and bot implementation from the utils package use the one on [this repository](https://github.com/larkox/mattermost-plugin-utils)

Since the repository is not yet a mattermost repository, and we are considering moving this to the plugin api repository, we shall not merge yet. Nevertheless, only the imports will change once this is decided, so it is open to dev review.

#### Ticket Link
None


mickmister:
This is great! We have been wanting to make a common repo for plugin utils for a while. We have been working on https://github.com/mattermost/mattermost-plugin-api to improve the PluginAPI specifically, but I don't think that's the right place for the sort of functionality you import in this PR.

I think we should go through the different plugin repos and lump in a bunch general purpose functions into the mattermost-plugin-utils repo.

PR #126, opened 2020-05-07
Implement autorespond feature

### LINKED ISSUE #108: The bot should reply to any DMs sent to a person in a meeting, saying they are busy
Bot should auto-respond when:
- User's `activeEvents` are not empty
- and User's status is not currently set to `online`

The reason for the second condition, is because this means the user has manually set themselves to online during the meeting. The `online` status is retained until their next meeting.

There is an [open PR](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/73) that will be merged soon that introduces the [`activeEvents`](https://github.com/larkox/mattermost-plugin-mscalendar/blob/7b2a6fe8e6dacdae5ae9768d59e4b138945c9dd6/server/store/user_store.go#L42) property. This property can also be added in this PR before the other PR is merged. The merge conflicts will be trivial to solve, if any.

This can be accomplish using the [`MessageHasBeenPosted`](https://developers.mattermost.com/extend/plugins/server/reference/#Hooks.MessageHasBeenPosted) hook.

MatthewDorner:
#### Summary
- Use `MessageHasBeenPosted` hook to have bot post an Ephemeral response to DMs when recipient is in a meeting and their status is not set to `online`.
- Implement the `/mscalendar autorespond` command to set a custom autoresponse message
- Add `AutoRespond` boolean and `AutoRespondMessage` string settings to the user in KV store. Add the `AutoRespond` boolean setting to the settings panel and welcome flow.

I'll add tests in the next couple days.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/108


MatthewDorner:
Some screenshots @asaadmahmood.

Slash command:
autorespond command
Slash command response:
autorespond command result
Bot auto-responding to DM:
autorespond response
Settings panel:
auto respond setting panel
Welcome flow:
welcome autorespond
autorespond welcome flow

If the user enables the autorespond setting, but doesn't set their message with the `/mscalendar autorespond` command, it will fall back to the default message, "This user is currently in a meeting.".

larkox:
@MatthewDorner let's wait for Asaad's feedback, but I think it looks great. The only thing I am missing is some information on how to change the autorespond text from the settings panel.
What do you think about adding a "read only setting" (the type should already exist) with information on how to change it and the current value?

aaronrothschild:
> @MatthewDorner let's wait for Asaad's feedback, but I think it looks great. 

I second this!  Looks great!

MatthewDorner:
@larkox I will add the read-only settings panel item next.

MatthewDorner:
Added read-only setting:
read only setting

larkox:
Thank you very much for the changes. I wanted to check if you had one case in mind when you implemented this (I see nothing in the code about this, but I may have missed something).

Here the example:
1. User A and User B are in a meeting with autorespond on.
2. User A writes to User B.
3. User B autoresponder writes to User A.
4. User A autoresponder writes to User B.
5. Infinite loop between steps 3 and 4.

My idea would be to store to which users you have answered since the last time you were "free". Then, every time you become free (this can be checked on the availability job) clean this list of users. Does it make sense?

server/mscalendar/auto_respond_test.go
```diff
@@ -0,0 +1,167 @@
...
+			if (tc.recipientStatusString != model.STATUS_ONLINE) {
+				mockStore.EXPECT().GetSetting("mattermost_user_recipient_id", store.AutoRespondSettingID).Return(tc.autoRespondSetting, nil)
+
+				if (tc.autoRespondSetting.(bool) && len(tc.recipientActiveEvents) > 0) {
+					mockStore.EXPECT().GetSetting("mattermost_user_recipient_id", store.AutoRespondMessageSettingID).Return(tc.autoRespondMessageSetting, nil)
+
+					if (tc.autoRespondMessageSetting != "") {
+						mockPoster.EXPECT().Ephemeral("mattermost_user_sender_id", "mattermost_post_channel_id", "Hello, I'm in a meeting and will respond to your message as soon as I'm free.")
+					} else {
+						mockPoster.EXPECT().Ephemeral("mattermost_user_sender_id", "mattermost_post_channel_id", "This user is currently in a meeting.")
```
mickmister, line 152: Instead of having this conditional logic in the tests (as we should not be duplicating the actual code's logic in the test), can we have a `expectedDMMessage` in the test case data? If the value is `""` then we can assume that the DM was not supposed to be sent.

server/utils/pluginapi/api.go
```diff
@@ -21,8 +21,28 @@ func New(api plugin.API) *API {
 	}
 }
 
+func (a *API) GetMattermostChannel(mattermostChannelID string) (*model.Channel, error) {
```
mickmister, line 24: This seems a bit verbose to call the var `mattermostChannelID`. The reason we do it with `mattermostUserID` is to distinguish between a local user and a remote user ID. But we do not have the concept of a remote channel.

```suggestion
func (a *API) GetMattermostChannel(channelID string) (*model.Channel, error) {
```

mickmister (changes requested):
Thanks for implementing this @MatthewDorner! I have a few requests regarding style and test case implementation.

MatthewDorner:
@larkox The response is an Ephemeral post by the bot, not the recipient's user. However, there is nothing in my code that would prevent this infinite loop IF the `MessageHasBeenPosted` hook was called for the bot response. I did test this case, and the hook does not get called for the response.

If you think I should add another check to explicitly check for a bot post, I can do that.

asaadmahmood (changes requested):
Also thinking of having a textfield above the yes/no of the autorespond setting, which allows users to enter the text and then press yes/no.

Thoughts @aaronrothschild @larkox @mickmister

larkox:
@asaadmahmood 
Text fields are not supported in slack attachments, so that would not be possible to implement in the current environment. I am working on an alternative, but it is still in progress. You can take a look here: https://community.mattermost.com/core/pl/n8fjqh6zppb9prkg44y4fh4bnh

larkox:
@MatthewDorner If you have made the tests and it works fine, I am OK with it.
@DHaussermann heads up on this as a test step, specially on main product releases, in case the hook behaviour changes for some reason. Context:

> Here the example:
> 
> 1. User A and User B are in a meeting with autorespond on.
> 2. User A writes to User B.
> 3. User B autoresponder writes to User A.
> 4. User A autoresponder writes to User B.
> 5. Infinite loop between steps 3 and 4.

asaadmahmood:
Okay, then looks good to me as of now.
I don't think the confirmation thing is very useful, as we can just tell them what their new status was set to, and they can just update it if they don't like it.
I would instead even think of most of these settings to behave like:

`/mscalendar autorespond` and then a user presses `enter`, a modal opens up with one textfield allowing them to set their autorespond status. They can set it there and click Done, and it's set.
But I'm assuming that modals don't work well on mobile?

mickmister:
> But I'm assuming that modals don't work well on mobile?

@asaadmahmood We can use an [Interactive Dialog](https://developers.mattermost.com/extend/plugins/server/reference/#API.OpenInteractiveDialog) to let the user input some [free text](https://docs.mattermost.com/developer/interactive-dialogs.html#textarea-elements). I think it would work fairly well and not require too much code.

MatthewDorner:
> @MatthewDorner If you have made the tests and it works fine, I am OK with it.
> @DHaussermann heads up on this as a test step, specially on main product releases, in case the hook behaviour changes for some reason. Context:
> 
> > Here the example:
> > 
> > 1. User A and User B are in a meeting with autorespond on.
> > 2. User A writes to User B.
> > 3. User B autoresponder writes to User A.
> > 4. User A autoresponder writes to User B.
> > 5. Infinite loop between steps 3 and 4.

If this happened, it probably wouldn't alternate between the two users' autoresponders but it would just be the same user autoresponding over and over. This is because the code I wrote only determines the recipient by checking if the id is not equal to the id on the incoming post (and assumes that a DM channel will only have 2 users, which I believe is true otherwise the type would be `CHANNEL_GROUP`). If the hook ran for the bot, it would just run through the `range` and assign `storedRecipient` both times. So this also means it could occur if only one user had autorespond turned on.

larkox:
> @asaadmahmood We can use an Interactive Dialog to let the user input some free text. I think it would work fairly well and not require too much code.

@MatthewDorner Would you mind exploring this option? The behaviour would be:
`/mscalendar autorespond message` sets message as the autorespond
`/mscalendar autorespond` opens the interactive dialog.

If you feel it is out of the scope of this PR, we can add it as another HW ticket, but it would be great if you could take a look.

MatthewDorner:
Yeah, I can. So, there would be two ways to set the message? What should the `/mscalendar autorespond` command response say in both cases? Just have it return an empty string in the case where it opens the dialog?

Also, I used the language "autorespond message" and "auto-response message" in some places, and "Auto Respond" as the title in the settings/welcome flow. I guess I should change them to all be the same, any preference?

larkox:
@MatthewDorner `/mscalendar autorespond message` should work as it is working now. The one with the dialog should not return any message (an empty string might do the trick, but right now I am not 100% sure).

Regarding naming, I think we should be consistent yes. Me not being native English speaker, I have no preference over one or the other.

MatthewDorner:
Added interactive dialog. `/mscalendar autorespond` opens the dialog. There is no response to the slash command itself, but if the message is successfully set through the dialog, I have it make an Ephemeral post with the same "Autorespond message changed to..." as I have for the response to `/mscalendar autorespond <message>`. `/mscalendar autorespond <message>` works the same as before.

I went with "Auto Respond" for titles, and "autorespond" inside sentences. No real basis for this other than it sounds good to me, let me know if I should change.

One question, when I create the `OpenDialogRequest`, I leave `CallbackId` and `State` as empty strings. It seems to work fine, but should I put something there?

autorespond dialog

server/mscalendar/auto_respond.go
```diff
@@ -0,0 +1,92 @@
...
+
+	message, ok := autoRespondMessage.(string)
+	if !ok {
+		return errors.Errorf("Error retrieving setting: %s", store.AutoRespondMessageSettingID)
+	}
+	if message == "" {
+		message = defaultMessage
+	}
+
+	m.Poster.Ephemeral(post.UserId, post.ChannelId, message)
```
larkox, line 82: What do you think about "impersonating the user"? That is, instead of posting as a bot, posting as the actual user that is away.

The current poster cannot do that, but should be easy to add a "PostDMAs" function. The upside of this is that the current server implementation does not add notifications when a "third user" (i.e. the bot) writes in a direct channel.

The notification may not be a huge concern here, since it will answer just after you write something, but just wondering on your view.

(This too is more to open the discussion than a proper change request)

larkox: I also wonder which one would be best, an ephemeral message or a lasting message. Not sure about this.

larkox (changes requested):
Thank you very much for the contribution! A few changes, nitpicks and questions.

server/mscalendar/welcome_flow.go
```diff
@@ -79,6 +79,17 @@ func (wf *welcomeFlow) makeSteps() {
...
 		Title:   "Daily Summary",
 		Message: "Remember that you can set-up a daily summary by typing `/mscalendar summary time 8:00AM`.",
+	}, &flow.SimpleStep{
+		Title:                "Auto Respond",
+		Message:              "Do you want to enable an automatic bot response to DMs when you are in a meeting?",
+		PropertyName:         store.AutoRespondPropertyName,
+		TrueButtonMessage:    "Yes - I would like to enable an automatic response",
+		FalseButtonMessage:   "No - Do not enable",
+		TrueResponseMessage:  "Great, the bot will automatically respond to DMs when you are in a meeting.",
+		FalseResponseMessage: "Great, automatic response will not be enabled.",
```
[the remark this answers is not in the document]
MatthewDorner: Yes, I think it should skip the empty step.

server/mscalendar/auto_respond.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	message, ok := autoRespondMessage.(string)
+	if !ok {
+		return errors.Errorf("Error retrieving setting: %s", store.AutoRespondMessageSettingID)
+	}
+	if message == "" {
+		message = defaultMessage
+	}
```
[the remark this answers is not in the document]
MatthewDorner: Since the user didn't set the default message to begin with, I don't think they'd mind if it was changed. Saving the message in the database would make it more difficult if, for some reason, we did want to change the default message for already-connected users.

server/mscalendar/auto_respond.go
MatthewDorner: I like using the Ephemeral bot post, as it makes it obvious that the user is not actually responding. And using a lasting post could clutter the conversation.

server/mscalendar/auto_respond.go
```diff
@@ -0,0 +1,92 @@
...
+
+	recipientStatus, err := m.PluginAPI.GetMattermostUserStatus(storedRecipient.MattermostUserID)
+	if err != nil {
+		return err
+	}
+	if recipientStatus.Status == model.STATUS_ONLINE {
+		return nil
+	}
+
+	autoRespond, err := m.Store.GetSetting(storedRecipient.MattermostUserID, store.AutoRespondSettingID)
```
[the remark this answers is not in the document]
MatthewDorner: Thanks, didn't even occur to me.

server/mscalendar/auto_respond.go
```diff
@@ -0,0 +1,92 @@
...
+	usersInChannel, err := m.PluginAPI.GetMattermostUsersInChannel(post.ChannelId, model.CHANNEL_SORT_BY_USERNAME, 0, 2)
+	if err != nil {
+		return err
+	}
+
+	var storedRecipient *store.User
+	for _, u := range usersInChannel {
+		storedUser, _ := m.Store.LoadUser(u.Id)
+		if u.Id != post.UserId {
+			storedRecipient = storedUser
```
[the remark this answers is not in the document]
MatthewDorner: yeah, fixed now.

server/mscalendar/auto_respond.go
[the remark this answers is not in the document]
larkox: @aaronrothschild any thoughts on this?

server/command/auto_respond.go
```diff
@@ -0,0 +1,45 @@
...
+						Name:        "auto_respond",
+						Type:        "text",
+						Placeholder: "Enter an autorespond message.",
+					},
+				},
+				SubmitLabel:    "Submit",
+				NotifyOnCancel: false,
+				State:          "",
+			},
+		}
```
MatthewDorner, line 32: Here's the `OpenDialogRequest` I mentioned earlier. Should I put anything for `CallbackId`and `State`?

mickmister: For the purposes of this dialog, I think we can leave them blank. @hanzei Do you have any insight on if these need to be used here?

server/mscalendar/auto_respond_test.go
```diff
@@ -0,0 +1,169 @@
...
+			mockPluginAPI.EXPECT().GetMattermostUsersInChannel("mattermost_post_channel_id", model.CHANNEL_SORT_BY_USERNAME, 0, 2).Return([]*model.User{mattermostUserSender, mattermostUserRecipient}, nil)
+
+			mockStore.EXPECT().LoadUser("mattermost_user_sender_id").Return(nil, errors.New("user not found"))
+			mockStore.EXPECT().LoadUser("mattermost_user_recipient_id").Return(storedRecipient, nil)
+
+			if tc.autoRespondSetting && len(tc.recipientActiveEvents) > 0 {
+				mockPluginAPI.EXPECT().GetMattermostUserStatus("mattermost_user_recipient_id").Return(recipientStatus, nil)
+			}
+
+			if tc.recipientStatusString != model.STATUS_ONLINE && tc.expectedDMMessage != "" {
```
mickmister, line 155: This way we are not assuming any logic of how the source code works

```suggestion
			if tc.expectedDMMessage != "" {
```

server/api/auto_respond.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/mscalendar"
+)
+
+func (api *api) autoRespond(w http.ResponseWriter, req *http.Request) {
```
mickmister, line 15: I think the method name should be more descriptive like this. Then the API route can be `set-autorespond-message`.

```suggestion
func (api *api) setAutorespondMessage(w http.ResponseWriter, req *http.Request) {
```

mickmister (changes requested):
Just a few more requests from me. Looking good otherwise :+1:

Also, I'm unsure if we are going with:

- `autorespond` vs `autoRespond`
- `Autorespond` vs `AutoRespond` vs `Auto Respond` vs `Auto-Respond`

We may not need to be consistent throughout, because in some contexts it seems to makes more sense to use one over the others. Just thought I'd point this out. Naming things is deemed one of the most difficult things in software.

server/command/auto_respond.go
```diff
@@ -3,19 +3,39 @@ package command
...
+			URL:       c.Config.PluginURL + config.PathDialog + config.PathAutoRespond,
+			Dialog: model.Dialog{
+				CallbackId: "",
+				Title:      "Auto Respond Message",
+				Elements: []model.DialogElement{
+					{
+						DisplayName: "Autorespond Message",
+						Name:        "auto_respond",
+						Type:        "text",
+						Placeholder: "Enter an autorespond message.",
```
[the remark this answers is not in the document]
MatthewDorner: This dialog won't allow users to submit a blank value.

server/command/auto_respond.go
```diff
@@ -3,19 +3,39 @@ package command
...
+		request := model.OpenDialogRequest{
+			TriggerId: c.Args.TriggerId,
+			URL:       c.Config.PluginURL + config.PathDialog + config.PathAutoRespond,
+			Dialog: model.Dialog{
+				CallbackId: "",
+				Title:      "Auto Respond Message",
+				Elements: []model.DialogElement{
+					{
+						DisplayName: "Autorespond Message",
+						Name:        "auto_respond",
```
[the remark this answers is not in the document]
MatthewDorner: I agree with the suggested change for the dialog field above. Regarding variable names, I think they are already consistent. They're all `autoRespond`, treating it as two words.

As far as text being displayed to the user, I guess the reason I felt I should go with "autorespond message" is that "auto respond message" feels ambiguous between (auto respond) message and auto (respond message), but I think the grammatically correct thing to do in such cases is hyphenat.

So, for the Welcome/Setting titles or places where it's just "auto respond", it can be two separate words, but when it's "auto-respond message", it can be hyphenated. Does that sound right?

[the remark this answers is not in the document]
mickmister: Sounds good to me :+1:

server/command/auto_respond.go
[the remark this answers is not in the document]
mickmister: Even if the form cannot be submitted with a blank value, I think it's the right placeholder to show in the form since it's the default value.

server/command/auto_respond.go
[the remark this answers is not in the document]
MatthewDorner: OK, so it will have "Auto Respond" as a title, "Auto-Respond Message" as a title (according to https://blog.apastyle.org/apastyle/2012/03/title-case-and-sentence-case-capitalization-in-apa-style.html), "auto-respond message" in a sentence, "Auto-respond message" at the beginning of a sentence (https://ontariotraining.net/writing-style-capitalization-of-hyphenated-words%E2%80%A8%E2%80%A8/).

server/api/auto_respond.go
MatthewDorner: To keep in convention with other naming ("auto" and "respond" being separate words), the API route would be `set-auto-respond-message`, the method name would be `setAutoRespondMessage`, and the constant in `config/const.go` would be `PathSetAutoRespondMessage`, but those are getting lengthy.

mickmister (approved):
LGTM :tada:

server/store/setting_store.go
```diff
@@ -8,9 +8,11 @@ import (
 const (
 	UpdateStatusSettingID               = "update_status"
 	GetConfirmationSettingID            = "get_confirmation"
-	ReceiveNotificationsDuringMeetingID = "receive_notification"
 	ReceiveRemindersSettingID           = "get_reminders"
 	DailySummarySettingID               = "summary_setting"
+	ReceiveNotificationsDuringMeetingID = "receive_notification"
```
larkox, line 13: nit: would you mind putting this line were it was before? Just to clean the diff.

larkox (approved):
LGTM. Just one tiny nit.

larkox:
Ping @asaadmahmood and @aaronrothschild . This has already passed Dev Review, in case you want to add any other comment before we pass to QA Review.

asaadmahmood:
@larkox Based on the last screenshot i see, we should get rid of the label above the input, as the header already mentions that. But other than that, looks good.
image

MatthewDorner:
@asaadmahmood Is this OK? (I changed the wording based on decision made in https://github.com/mattermost/mattermost-plugin-mscalendar/pull/126#discussion_r429437075)
remove label

asaadmahmood:
@MatthewDorner Can we remove the red star in the label above the input?

MatthewDorner:
The field can either be required or optional. The red star appears if the field is required, otherwise it looks like this (it will also behave differently if it's optional, allowing the user to submit with it empty). It seems like the dialog is designed to expect fields to have labels.
optional autorespond dialog

larkox:
@MatthewDorner What about changing the title to the title of the plugin (Microsoft Calendar) and then the title of the field as "Auto-respond message" (or whichever is the standard we decided to use on the naming)?

@asaadmahmood Do you see this as a good idea?

asaadmahmood:
@larkox If its a limitation, then yeah, that makes sense.

larkox:
@asaadmahmood To give you some context:
We are using an "Interactive Dialog", a pre-made dialog from the web app that can be configured to have several different types of inputs (text-boxes, options dropdowns, checkmarks...). It is usually used with several fields, so the original implementation was not considering just having title, and one text box with nothing around it.

Since this is something pre-made on the webapp, changing it would require changes on the webapp and being tied to certain version of the webapp.

We could create a new component ourselves, and try to make it look similar, but any effort in that sense would require much more work, and potentially would end up being less integrated.

If you want to know more about interactive dialogs, how can they be used, and their limitations, please let me know.

asaadmahmood:
@larkox Yup, I am a bit aware of what Interactive DIalogs are and why we use them. So yeah, if not having a label is something we can't do right now, then we can go with what you suggested.

asaadmahmood (approved):

DHaussermann (approved):
Tested and passed
I have tested this feature and everything seems to be working as expected.
- Autorespond  does nothing when disabled 
- Autorespond  does nothing when enabled and user says they wish to receive notifications during  events (set to away by MS Cal plugin instead of DND)
- Autorespond  responds automatically with an ephemeral post that contains the configured text when user status is updated to busy
- Autorespond stops responding if even has ended and user status is reverted 
- Autorespond stops responding if user manually modifies status to Online during an event 
- Autorespond stops responding if you're in an event as busy and disable it from the settings panel 
- Autorespond starts responding if enabled from the settings panel after a meeting has started  
- Current response text for Autorespond is correctly shown in settings panel
- Autorespond is functional if enabled and no message has been provider by the user. ("This user is currently in meeting")
- As the response in an ephemeral post and not a DM from the user, it will not create a feedback loop if 1 meeting attendee attempts to DM an other :) 
LGTM!

Added test cases to release testing.
Huge thanks @MatthewDorner for this enhancement. This is great!
Sorry for the long delay in testing this.

larkox:
@MatthewDorner Since we changed again the visibility to public, you will have to create a new fork for this repo, since the links seem to have broken again :sweat_smile:
Nevertheless, if you follow the steps I pasted before with this new fork, it should work.

MatthewDorner:
Sorry, I haven't been checking on GitHub. I'll try to do this.

Issue #123, opened 2020-05-07
Remove stored `EventSubscriptionID` when we know it is stale


mickmister:
I have a stale event subscription in my local env, which is expected because I just recently had the renew job merged/running. But I receive this error each run of the renew job: `Error renewing subscription: 404 Not Found: {"error":{"code":"ResourceNotFound","message":"The object was not found."}}`.

I'm thinking we should remove the local subscription when this happens, especially if we detect `Not Found` to be in the error string.

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/b4ddb3baee0dffd062c10f5a7bbf1114f63994cd/server/mscalendar/subscription.go#L82-L85

PR #131, opened 2020-05-13
When trying to renew an expired subscription, create new subscription instead

mickmister:
#### Summary

When the renew job runs and the user's subscription has expired, the call to `client.RenewSubscription` fails. This PR makes it so when this occurs, the local subscription is removed and a new one is created.

This happened to me as I had an expired subscription on one of my local MM accounts before the renew job was implemented.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/123


mickmister:
@DHaussermann A customer is running into an issue similar to this and it reminded me that the error this PR solves may occur as well. I think we should raise the priority of this fix. What are your thoughts?

DHaussermann (approved):
Tested and passed. This issue seems resolved. 
- Setup MM Server and connected to MS Calendar App
- Created a subscription and noted expiry date
- Shut-down MM server for 5+ days to ensure the expiry date had lapsed
- Started up MM server and logged in as test user
- Ran `/mscalendar subscribe list`
Valid subscription was in place with an expiry in the future.

LGTM!

Issue #115, opened 2020-05-05
No explanation of the system console config values related to scheduled jobs


mickmister:

mickmister:
Closing in favor of https://github.com/mattermost/mattermost-plugin-mscalendar/issues/144

PR #137, opened 2020-05-20
Explain jobs and enable jobs by default

mickmister:
#### Summary

This PR explains the two scheduled jobs that are configurable to be enabled/disabled in the system console.

It also enables the jobs by default. Having an explicit step in the config to enable them is another option.

The additions can be found at the end of the README here https://github.com/mattermost/mattermost-plugin-mscalendar/blob/035e8606ab8f380595086bc9ebc4c723b1c2d2a1/README.md

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/115

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,3 +101,16 @@ Copy the `Client ID` and `Tenant ID` from the Azure portal
 - `tenantID` - copy from Azure App
 - `clientID` - copy from Azure App
 - `Client Secret` - copy from Azure App (Generated in `Certificates & secrets`, earlier in these instructions)
+
+**Scheduled Jobs**
+
+There are two scheduled jobs than can be disabled from the plugin settings in the system console.
+
+The Status Sync job runs every 5 minutes and is responsible for:
+
+- Applying status changes to Mattermost users when they are current availability on their calendar
+- Delivering upcoming event reminders to users
+
+The Daily Summary job runs every 15 minutes and is responsible for:
+
+- Delivering each user's calendar summary for the day, based on their settings of when they want to receive the summary
--- a/plugin.json
+++ b/plugin.json
@@ -84,14 +84,14 @@
                 "display_name": "Enable User Status Sync Job",
                 "type": "bool",
                 "help_text": "When enabled, a Mattermost user's status will automatically update based on their Microsoft Calendar availability. This runs every 5 minutes.",
-                "default": false
+                "default": true
             },
             {
                 "key": "EnableDailySummary",
                 "display_name": "Enable Daily Summary",
                 "type": "bool",
                 "help_text": "When enabled, Mattermost users will a receive a daily summary of their scheduled events in Microsoft Calendar. Users can choose when the summary will be sent to them.",
-                "default": false
+                "default": true
             }
         ]
     }


larkox:
Two considerations:
1. The subscription renewal is a different job, or was somehow piggybacked in one of these jobs?
2. If by system configuration you can disable the status sync job, does it make sense to ask the user if they want to sync their status? Similar with daily summary and so on. If not, we should "remove "the flow steps, settings and slash commands relevant to each setting if they are disabled.

mickmister:
> The subscription renewal is a different job, or was somehow piggybacked in one of these jobs?

The renewal job is a different job, and cannot be disabled. I assumed @levb had set this up by design. 0/5 on making a config value for it

> If by system configuration you can disable the status sync job, does it make sense to ask the user if they want to sync their status? Similar with daily summary and so on. If not, we should "remove "the flow steps, settings and slash commands relevant to each setting if they are disabled.

The jobs can be enabled/disabled at any point, even temporarily due to some circumstance. I wouldn't want the flow to exclude something that the user would otherwise have guidance to set up, if it is a temporary disabling. I will defer to @aaronrothschild and @asaadmahmood here

larkox:
> The jobs can be enabled/disabled at any point, even temporarily due to some circumstance. I wouldn't want the flow to exclude something that the user would otherwise have guidance to set up, if it is a temporary disabling. I will defer to @aaronrothschild and @asaadmahmood here

My point would be the exact opposite. I do not want to show the user settings for something that they are not receiving. The user will setup the daily summary, and the daily summary will never come.

Maybe, we could consider a messaging system that DM all connected users when one of the jobs is enabled/disabled to let them know that those features are available/unavailable.

Another solution would be to remove the possibility to enable or disable the jobs. Is there any strong reason to have those configurable?

asaadmahmood:
@mickmister @larkox Can I please get some content what "jobs" means in this context? and then which two jobs are we talking about here?

mickmister:
> Another solution would be to remove the possibility to enable or disable the jobs. Is there any strong reason to have those configurable?

@larkox I think it is because the job code has not been battle-tested in production before, so we want a way for the admins to disable them if necessary. However, the plugin isn't all that useful without the jobs so maybe we should get rid of the config vars.

README.md
```diff
@@ -101,3 +101,16 @@ Copy the `Client ID` and `Tenant ID` from the Azure portal
...
+There are two scheduled jobs than can be disabled from the plugin settings in the system console.
+
+The Status Sync job runs every 5 minutes and is responsible for:
+
+- Applying status changes to Mattermost users when they are current availability on their calendar
+- Delivering upcoming event reminders to users
+
+The Daily Summary job runs every 15 minutes and is responsible for:
+
+- Delivering each user's calendar summary for the day, based on their settings of when they want to receive the summary
```
mickmister, line 116: @asaadmahmood Here's an excerpt from the PR explaining the two jobs

mickmister: The jobs are functions written on the server-side of the plugin to run at different intervals, depending on the purpose of the job

mickmister:
Closing in favor of removing the job configs https://github.com/mattermost/mattermost-plugin-mscalendar/pull/145

PR #143, opened 2020-05-22
[GH-14] Add at-rest encryption for OAuth2 token

### LINKED ISSUE #14: Add at-rest encryption for OAuth2 token
Presently, the POC code stores the OAuth token(s) unencrypted. It is a best practice to have it encrypted when stored persistently, so that the encryption keys may (eventually) be stored separately.

This improvement is currently pending to the Zoom plugin, can copy from there, https://github.com/mattermost/mattermost-plugin-zoom/pull/52/files, search for `EncryptionKey`. Note, `TokenEncryptionKey` is likely a more intention-revealing name.

larkox:
#### Summary
Add at-rest encryption for OAuth2 token
#### Ticket Link
Fix #14


levb:
Do we really need to QA this one, other than the general Oauth2/use regression?

larkox:
Use and regression is all I can think about. Proper behavior on upgrade could also be necessary, but since we still do not have any release out, I think it is not necessary.

DHaussermann:
@larkox when I deploy this, I can no longer run connect or disconnect command. Error log shows 
```
{"level":"debug","ts":1590523742.088871,"caller":"web/handlers.go:89","msg":"Received HTTP request","method":"POST","url":"/api/v4/commands/execute","request_id":"bfitqmz44ifkbem5q7tkzai99w"}
{"level":"error","ts":1590523742.0895622,"caller":"mlog/sugar.go:23","msg":"token encryption key not generated","plugin_id":"com.mattermost.mscalendar"}
{"level":"error","ts":1590523742.089841,"caller":"mlog/log.go:175","msg":"Unable to execute command.","path":"/api/v4/commands/execute","request_id":"bfitqmz44ifkbem5q7tkzai99w","ip_addr":"74.15.49.102","user_id":"ri814hfoubngxp8e838raoxxsw","method":"POST","err_where":"mscalendarplugin.ExecuteCommand","http_code":500,"err_details":"token encryption key not generated"}
```
Please advise on any next steps for investigation.

DHaussermann:
@larkox disregard previous comment `At Rest Token Encryption Key` was missing. 
Regarding the testing: 
- Tested disconnect and connect with multiple users. 
- Functional testing found nor regreassions
However for the test user where I watched the KV store I did not see a change in `access_token`
is there a way to validate the current value is properly encrypted?

larkox:
This should do the trick:
https://play.golang.org/p/67MSocnXQn9
I put there the decryption code. You you have to fill the key and the "message to decrypt" and hit run. If it is something that has not been encrypted before, it should not work. Therefore, if what is stored in the KVStore passes the decryption, it should be fine. If you need any more guidance, feel free to ask.

Issue #144, opened 2020-05-26
Remove scheduled job config options


mickmister:
We are removing the plugin setting configs for the scheduled jobs, as having them disabled is essentially the same thing as disabling the plugin as far as functionality.

Issue #135, opened 2020-05-20
Implement verbose logging for HA jobs


mickmister:
In order to be able to trace any issues with HA jobs, we should log actions before they are attempted, so if something locks up, we can determine which operation got stuck.

PR #145, opened 2020-05-26
Remove job configs

mickmister:
#### Summary

Scheduled jobs are no longer able to be disabled. They are still synchronized via the cluster job locks.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/144
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/135


server/jobs/job_manager.go
```diff
@@ -58,47 +57,20 @@ func NewJobManager(papi cluster.JobPluginAPI, env mscalendar.Env) *JobManager {
...
 // Close deactivates all active jobs. It is called in the plugin hook OnDeactivate.
 func (jm *JobManager) Close() error {
+	jm.env.Logger.Debugf("Deactivating all jobs due to plugin deactivation.")
 	jm.activeJobs.Range(func(k interface{}, v interface{}) bool {
 		job := v.(*activeJob)
 		err := jm.deactivateJob(job.RegisteredJob)
 		if err != nil {
-			jm.env.Logger.Debugf("Failed to deactivate job: %v", err)
+			jm.env.Logger.Warnf("Failed to deactivate %s job: %v", job.id, err)
 		}
```
mickmister, line 74: @lieut-data We are experiencing an issue testing the cluster jobs in HA. We are testing with a fresh cloud server with `--size miniHA`, which spins up a cluster with two nodes.

I will be referring to the jobs as a singular "job" (`cluster.Job`), as all the jobs are showing the same behavior. The job activates properly on plugin activation. The successful activation message is logged by both nodes by the `jm.activateJob` method. The job runs its first execution with no issues. However, the job does not run again when its configured interval comes around.

Disabling the plugin results in the call to `job.Close()` to hang. Tracing the logs provided by this PR, the `Deactivating all jobs due to plugin deactivation` message is logged once (due to the node blocking on disable, and not allowing the second node to begin disabling), and does not log the debug message in `jm.deactivateJob`, nor the warning in this function. Therefore the `job.Close()` method must be blocking.

I'm not sure where to begin with debugging this. I'd like to debug locally, but I haven't set up an HA environment on my local machine. How do you advise moving forward with this?

mickmister: To be clear, this behavior was occurring before this PR. Part of the reason for this PR is to simplify the logic, to minimize the surface area for errors like the ones described above to occur.

DHaussermann:
Tested and passed
- Tested this on an HA server
- Jobs are working as expected
- No locks found when disabling plugin
LGTM!

server/jobs/job_manager.go
lieut-data: (For anyone following this thread, see https://community-daily.mattermost.com/core/pl/auyr1cd1q7nu9ytmtb9a5fod8a for the resolution.)

PR #151, opened 2020-05-28
Add EE license check

larkox:
#### Summary
Add license check on plugin activate.

IMPORTANT: This require v5.24.0 of mattermost-server which is not yet released. Therefore, the package cannot be imported, and go.sum has not been properly updated. Probably this PR needs an update when v5.24.0 is in place.

#### Ticket Link
None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/jarcoal/httpmock v1.0.4
 	github.com/mattermost/mattermost-plugin-api v0.0.9
-	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200313113657-e2883bfe5f37
+	github.com/mattermost/mattermost-server/v5 v5.24.0
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/jarcoal/httpmock v1.0.4
 	github.com/mattermost/mattermost-plugin-api v0.0.9
-	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200313113657-e2883bfe5f37
+	github.com/mattermost/mattermost-server/v5 v5.24.0
```
hanzei, line 10: Could some commit on `master` be used instead?

larkox: This was discussed on a different conversation. We will point to master, merge this, and whenever 5.24 is addressable, we will update the plugin to address the correct server version.

larkox:
@aaronrothschild according to [this conversation](https://community.mattermost.com/core/pl/ys5icsq3utbcjx6abkjbcfo14a) we discussed about checking for LDAP until we start using 5.24. Should that check be done in this PR?

PR #152, opened 2020-05-28
Add Telemetry to MSCalendar

larkox:
#### Summary
First draft of telemetry package, implemented on MSCalendar. The objective is to have this package extracted into plugins-api repo.

Rudder specific information may be included in the code here. Once the information is in place, this should work.

#### Ticket Link
None.


server/mscalendar/mscalendar.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/store"
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/settingspanel"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/telemetry"
```
levb, line 14: 👍

server/utils/telemetry/rudder.go
```diff
@@ -0,0 +1,38 @@
...
+
+import rudder "github.com/rudderlabs/analytics-go"
+
+const (
+	rudderDataPlaneURL = ""
+	rudderWriteKey     = ""
+)
+
+func NewRudderClient() (Client, error) {
+	client, err := rudder.NewWithConfig(rudderWriteKey, rudderDataPlaneURL, rudder.Config{})
```
mickmister, line 11: 0/5 Both of this function and `NewRudderClientWithCredentials` are very small, but couldn't this function call `NewRudderClientWithCredentials` to simplify? Maybe also match the argument order passed to `NewWithConfig`?

server/mscalendar/mscalendar.go
hanzei: :100:

server/plugin/plugin.go
```diff
@@ -173,6 +174,12 @@ func (p *Plugin) OnConfigurationChange() (err error) {
 		if err != nil {
 			e.Logger.Errorf("Error updating job manager with config. err=%v", err)
 		}
+
+		telemetryClient, err := telemetry.NewRudderClient()
+		if err != nil {
+			e.Logger.Errorf("Cannot create telemetry client. err=%v", err)
```
hanzei, line 180: Should the code `return` here to ensure `e.Tracker` doesn't get overwritten?

server/utils/telemetry/rudder.go
```diff
@@ -0,0 +1,38 @@
+package telemetry
+
+import rudder "github.com/rudderlabs/analytics-go"
```
hanzei, line 3: Why `github.com/rudderlabs/analytics-go` is not part of `go.mod`?

server/utils/telemetry/rudder.go
```diff
@@ -0,0 +1,38 @@
...
+
+const (
+	rudderDataPlaneURL = ""
+	rudderWriteKey     = ""
+)
```
hanzei, line 8: What is the short and the mid term solution to get this information? Can we squeeze a new RPC API method into 5.24? Or hard code it given that it won't change?

larkox: The plan here is to have it hard coded. I still need to discuss with the analytics team to decide exactly what values we will use here, but that is the plan.

It is not my favorite choice, but make things incredibly simpler, and it seems it is an acceptable risk (NPS plugin is doing a similar thing).

server/utils/telemetry/rudder.go
larkox: My mistake, forgot to add it to the commit.

server/plugin/plugin.go
larkox: This actually can be done completely different. I originally created the client here, since I thought I needed more variables to create the client (that may change on configuration change). Nevertheless, since the client is always the same, we could declare the client somewhere else, and just write the tracker.

Regarding having the tracker getting overwritten, my original idea is that if there is an error with the creation of the client, the telemetryClient variable would be nil. Creating a tracker with a nil client will just create a tracker that doesn't track, making a silent error that admins can deal with (or not) without interfering with the plugin behaviour. We also want to update the tracker either way, because we might have changed the "EnableDiagnostics" variable, and we want to stop either way the tracking. If we do not update the Tracker, we will be using the "old client" to track things even if the EnableDiagnostics have been changed to false.

hanzei: Got it. Thanks for the detailed explanation :+1:

server/plugin/plugin.go
```diff
@@ -73,10 +74,22 @@ func (p *Plugin) OnActivate() error {
 	}
 
 	command.Register(p.API.RegisterCommand)
+
+	p.telemetryClient, err = telemetry.NewRudderClient()
+	if err != nil {
+		p.env.bot.Errorf("Cannot create telemetry client. err=%v", err)
+	}
 	return nil
```
hanzei, line 82: Nit:
```suggestion
	}
	
	return nil
```

server/plugin/plugin.go
```diff
@@ -73,10 +74,22 @@ func (p *Plugin) OnActivate() error {
 	}
 
 	command.Register(p.API.RegisterCommand)
+
+	p.telemetryClient, err = telemetry.NewRudderClient()
+	if err != nil {
+		p.env.bot.Errorf("Cannot create telemetry client. err=%v", err)
```
hanzei, line 80: If this fails. `p.telemetryClient` is `nil`. I'm not sure if it's worth to care about the complexity. 1/5 to just return with an error here.

larkox: I think we do not want the plugin not  to activate due to an outage of Rudder servers, or the plugin being isolated of the internet, or a firewall blocking non-api calls. If we were to return with an error here, cases like the ones I just said would make the plugin unusable.

hanzei: Good point, makes sense :+1: Does the code need an update to handle that case?

larkox: No, telemetryClient will be set to nil, and a when used to create a new tracker, a tracker with a nil client will just not track.

larkox: We could make that more obvious with a check like this:
```
if p.telemetryClient == nil {
   telemetry.NewTracker(nil, ....)
} else {
   telemetry.NewTracker(p.telemetryClient, ....)
}
```
But I think this is not the way to go.

DHaussermann:
Hi @larkox I've tried this a couple times and see some data coming back.  
I see `userDeauthenticated`, `automaticStatusUpdate`, and `dailySummarySent`

However, I do not see `welcomeFlowCompletion` and `userAuthenticated` I have connected a Mattermost user to MS Calendar and run through the login flow a couple times. I see no reason why these 2 events are not coming back.  
Is this something you're able to repro.

larkox:
@DHaussermann There was an issue indeed with those events. Should be fixed now.

Issue #155, opened 2020-05-29
Possible issue with msgraph tokens expiring


DHaussermann:
There may be an issue where the users MS Graph token are expiring. This may mean that the user needs to reconnect for the plugin to function normally. 

I saw this occur twice (with 2 different users) I have data setup and will update this issue with results if I see it occur again. I don't know that there are any functional implications. Testing is ongoing. 
![Disconnect](https://user-images.githubusercontent.com/35588783/83307454-bcbe1780-a1d2-11ea-9c5d-1b72c2a817aa.PNG)

- Deploy plugin build (I used master https://github.com/mattermost/mattermost-plugin-mscalendar/commit/b549fad795475f4f478f48000d6009a68cde6a64) to MM server
- As a test user, connect to MS with the plugin 
- Wait 1 full day
- Run `/mscalendar disconnect`
**Observed**:
```
(log WARN) failed to delete remote subscription 383446f3-2d0b-415f-a564-693926fd00ff. err=msgraph DeleteSubscription: 401 Unauthorized: {"error":{"code":"InvalidAuthenticationToken","message":"Access token has expired.","innerError":{"date":"2020-05-29T19:44:33","request-id":"9009bb5f-ab90-444d-b1e2-3d7405ce90ef"}}}
```
**Expected** : Disconnect normally

DHaussermann:
@mickmister this seems to be reproducible about 1 day or so after the user initially connected.

This time instead of trying to disconnect, I used `viewcal` the log shows 
```
{"level":"error","ts":1590871016.629838,"caller":"mlog/sugar.go:23","msg":"Command /mscalendar viewcal failed: msgraph GetMailboxSettings: 401 Unauthorized: {\"error\":{\"code\":\"InvalidAuthenticationToken\",\"message\":\"Access token has expired.\"}}","plugin_id":"com.mattermost.mscalendar"}
[...]

PR #157, opened 2020-06-01
Revert "[GH-14] Add at-rest encryption for OAuth2 token (#143)"

mickmister:
This reverts PR [#143 Add at-rest encryption for OAuth2 token](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/143)

#### Summary

With the OAuth encryption code active, the `expires` field in the stored OAuth token is always being set to 0001-01-01T00:00:00Z, when it should be set to time of `connect time + one hour`, something like `2020-06-01T00:14:55.338026-04:00`. This causes Go's OAuth library to think the token is always valid and does not need a refresh.

Since the OAuth tokens expire after one hour, the token stops working after an hour. And since the OAuth library thinks the token is valid, it will not refresh. So no operations work after an hour from connecting.

We've also determined that encrypting the OAuth token brings little to no value because of how the encryption key is stored. It can also make debugging a little more difficult due to not having direct access to the relevant data in the KV store.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/155

#### Test Steps

- Deploy two different builds on two different servers, one on master, one with this PR.
- Connect a user with both builds. After one hour, the master build should show the failures, but the reverted PR build will not have any issues.


mickmister:
Some discussion on Mattermost here https://community-daily.mattermost.com/core/pl/rdnno6ttqig9ddw8rfz87wwg9w

PR #158, opened 2020-06-01
Improve message on login flow for receiving notifications during meetings

larkox:
#### Summary
 Improve message on login flow for receiving notifications during meetings to minimize the text on the buttons and comply with mobile limitations

#### Ticket Link
None


larkox:
Since the changes are only texts, I think that after PM Review and CodeOwner review we can just merge without a proper dev review nor QA review.

mickmister:
I think we should somehow still convey _why_ we are giving them these two options. Maybe just a quick aside "Setting to Do Not Disturb will silence notifications." Even though that's probably already obvious, maybe it's not necessary.

Issue #45, opened 2020-03-04
Add missing commands to help text


mickmister:
#### Summary

The help text should be updated to include all existing commands. The help text and command autocomplete message needs to be updated whenever a new command is added. The autocomplete is updated in https://github.com/mattermost/mattermost-plugin-mscalendar/pull/38 .

Also, the help text should be implemented in the currently unused `getHelp()` function, which can then be referenced in `c.help`.

DHaussermann:
@aaronrothschild this is this something we can address for the beta? 
The help text also contains references to features that where not implemented.

aaronrothschild:
@DHaussermann Yes, let's address those before release then.  

@mickmister Can I make the changes to the help text in the code directly? (without doing all the stuff you mentioned above) :
- Remove any non-public commands that didn't make it to release
- Update the autocomplete text pane to reflect those changes

PR #161, opened 2020-06-02
Fix help text

mickmister:
#### Summary

Removed:

- info (mainly a debug command)
- showcals
- createcal
- deletecal
- createevent
- findmeetings

Added:

- disconnect
- settings
- summary

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/45

### DIFF
--- a/server/command/help.go
+++ b/server/command/help.go
@@ -10,16 +10,7 @@ import (
 )
 
 func (c *Command) help(parameters ...string) (string, bool, error) {
-	resp := fmt.Sprintf("Mattermost Microsoft Calendar plugin version: %s, "+
-		"[%s](https://github.com/mattermost/%s/commit/%s), built %s\n",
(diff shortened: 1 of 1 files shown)


mickmister:
I've taken out the build info from the help text, and instead included the `info` command in the help text. Here's the current state:

image

aaronrothschild:
> We conciser a future improvement where summary is shown with it's own section. For now I think this is good. The summary command are very discoverable as the user will see this list when they run it without any additional parameters.

@DHaussermann 0/5 on adding a new section, but it does make sense to show what is the "default" parameter if nothing is specified.

mickmister:
@aaronrothschild When nothing is specified, it shows help text for the command. I didn't want to mess with the help too much in order to keep this PR short and sweet.

PR #163, opened 2020-06-03
Update flow texts

### LINKED ISSUE #156: Clarify wording on login flow
I think there may a be a couple ways to improve the wording on the login flow

1. Change `Do you want to receive notifications when you receive a new event?` to be `Do you want to receive notifications when you are invited to an event?`
This may be more clear.

2. Change `Do you wish your Mattermost status to be automatically updated to be Do not disturb at the time of your Microsoft Calendar events?` to be `Do you wish your Mattermost status to be automatically updated at the time of your Microsoft Calendar events?` 
There is another question after to pick between away and DND. The question of the text should not assume the user will select DND. We could also go with more common language and use `Would you like your Mattermost status to be automatically updated at the time of your Microsoft Calendar events?`

larkox:
#### Summary
Update flow texts to be more user friendly.

#### Ticket Link
Fix #156

### DIFF
--- a/server/mscalendar/welcome_flow.go
+++ b/server/mscalendar/welcome_flow.go
@@ -53,7 +53,7 @@ func (wf *welcomeFlow) makeSteps() {
 	steps := []flow.Step{}
 	steps = append(steps, &flow.SimpleStep{
 		Title:                "Update Status",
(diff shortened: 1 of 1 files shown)


larkox:
@DHaussermann This is just some text changes. Let me know if we can skip QA Review.

Issue #148, opened 2020-05-28
Intermittent error where disconnect fails 


DHaussermann:
There is an issue I've seen a couple time where `mscalendar disconnect` fails. 
The debug show that the subscription is deleted but then the plugin crashes and the disconnect does not complete.
![DisconnectCrash](https://user-images.githubusercontent.com/35588783/83101892-58d80980-a081-11ea-96ec-2a988f269be5.PNG)

Not sure if it related but I see the following panic occur:
```
{"level":"debug","ts":1590641888.3242419,"caller":"plugin/hclog_adapter.go:53","msg":"panic: runtime error: invalid memory address or nil pointer dereference","plugin_id":"com.mattermost.mscalendar"}
{"level":"debug","ts":1590641888.3242915,"caller":"plugin/hclog_adapter.go:53","msg":"[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xd078c9]","plugin_id":"com.mattermost.mscalendar"}
{"level":"debug","ts":1590641888.3243175,"caller":"plugin/hclog_adapter.go:53","msg":"","plugin_id":"com.mattermost.mscalendar"}
{"level":"debug","ts":1590641888.32433,"caller":"plugin/hclog_adapter.go:53","msg":"goroutine 39 [running]:","plugin_id":"com.mattermost.mscalendar"}
```
Not sure if the log is helpful. Sorry I wasn't able to go back and get more from the log.

This happened on build https://github.com/mattermost/mattermost-plugin-mscalendar/pull/120/commits/364913f96f7f92186a68aac710c357308001bd24

After restarting the plugin and trying to disconnect the user again. This issue no longer reproduced.

larkox:
Panics usually give a whole stack trace of all the functions called to reach that panic. Not sure if they also appear in that view, but should be just the following lines (a few of them).

Usually, the shape of the lines is something like:
```
github.com/mattermost/mattermost-plugin-mscalendar/some/path.(objectname).functionname(gibberish)
github.com/mattermost/mattermost-plugin-mscalendar/some/path:lineNumber
...
```

If you have by any chance that stack trace, it would be really helpful to pinpoint the issue.

PR #165, opened 2020-06-03
Change dep initialization order to avoid nil Logger

mickmister:
#### Summary

This is the only potential change I could find that could solve the intermittent issue of the Logger not being assigned to `Env.Dependencies.Remote`.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/148

#### Analysis

The most recent panic Dylan has seen related to this is happening intermittently on subscription deletion in the remote client's `DeleteSubscription` method. I have not seen this happen myself. It seems that `Logger` assigned to the remote client is nil.

(Before this code change) Remote is assigned here. Logger is assigned at the bottom of this range.
https://github.com/mattermost/mattermost-plugin-mscalendar/blob/63e1af836a0dc2bffffb41f789743e053682e2d2/server/plugin/plugin.go#L129-L135

MakeClient creates a new client, and initializes it with a logger taken from a `Remote` instance.
https://github.com/mattermost/mattermost-plugin-mscalendar/blob/63e1af836a0dc2bffffb41f789743e053682e2d2/server/remote/msgraph/remote.go#L39-L49

Panic is happening on subscription deletion here. `c.Logger` must be nil. The full panic trace is not in the description of [the issue](https://github.com/mattermost/mattermost-plugin-mscalendar/issues/148), but Dylan and I looked at it together at the time.
https://github.com/mattermost/mattermost-plugin-mscalendar/blob/63e1af836a0dc2bffffb41f789743e053682e2d2/server/remote/msgraph/subscription.go#L49-L60

The first time we saw this occur was during `HandleWebhook`:
https://github.com/mattermost/mattermost-plugin-mscalendar/blob/63e1af836a0dc2bffffb41f789743e053682e2d2/server/remote/msgraph/handle_webhook.go#L29-L39

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -126,9 +126,9 @@ func (p *Plugin) OnConfigurationChange() (err error) {
 		e.Config.MattermostSiteHostname = mattermostURL.Hostname()
 		e.Config.PluginURL = pluginURL
 		e.Config.PluginURLPath = pluginURLPath
-		e.Dependencies.Remote = remote.Makers[msgraph.Kind](e.Config, e.Logger)
 
 		e.bot = e.bot.WithConfig(stored.BotConfig)
+		e.Dependencies.Remote = remote.Makers[msgraph.Kind](e.Config, e.bot)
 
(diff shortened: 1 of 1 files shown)


levb:
Blame me.

Issue #33, opened 2020-02-07
Build the auth flow to include admin consent


mickmister:
#### Summary

In order to make setup of the plugin as easy as possible, we should minimize the interaction in the Azure portal. By using the admin consent flow, we can grant the bot account the right permissions in a more automated way than having the admin manually select the permissions in the portal.

The process of setting up the auth flow can be found here: https://docs.microsoft.com/en-us/graph/auth-v2-service#3-get-administrator-consent

Delegated permissions:
- Calendars.ReadWrite
- Calendars.ReadWrite.Shared
- MailboxSettings.Read

Application permissions:
- Calendars.Read
- User.ReadAll
- MailboxSettings.Read

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21690

mickmister:
Closing due to incomplete requirements

PR #176, opened 2020-07-06
Implement admin consent flow

mickmister:
#### Summary

This was implemented right now due to a customer request. Here is the quote from them:

> At the moment User Consent permissions are allowed but in about a week or so, all applications will be required to move to Admin Consent only

This is coming from a Mattermost admin, relaying a message from their Azure admin. I originally took this as "Any application-level permissions must be approved via admin consent, and not in the Azure console", which makes sense.

But reading their message again, it is saying that ALL permissions must be application-level. This seems backwards to me, as I would think that it would be more desirable to have all of the permissions be user-delegated. I'm thinking that the message from the Azure admin may have been relayed incorrectly. I'm following up with the customer about this.

If all permissions were application-level, then actions like creating a meeting would not be processed by Microsoft properly.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/33


mickmister:
Closing due to incomplete requirements

PR #187, opened 2020-08-21
Update how to fetch diagnostic id to avoid panics

larkox:
#### Summary
Check if the diagnostic information is nil to avoid panics

#### Ticket Link
None


DHaussermann (approved):
Tested and passed
- Checked diagnostic flag in true and false state
- Plugin functions fine when diagnostic is false. Could not repro any crashes
- No issues found
LGTM!

PR #183, opened 2020-08-06
Add Fallback to Slack Attachments

larkox:
#### Summary
Add Fallback to Slack Attachments, so text is written in the mail sent on notifications and other clients that may not accept Slack Attachments.

#### Ticket Link
None

PR #199, opened 2020-11-23
Fix panic when Daily Summary is nil for a user

larkox:
#### Summary
There seems to be a panic in master that happens when the daily summary settings are set to nil. This should never happen, but maybe a weird status in the system is triggering this.

This PR makes sure the daily summary setting is set by adding a default in every point it is present.

Another solution could be remove the indirection and trust the go defaults.

#### Ticket Link
None

### DIFF
--- a/server/mscalendar/daily_summary.go
+++ b/server/mscalendar/daily_summary.go
@@ -58,6 +58,14 @@ func (m *mscalendar) SetDailySummaryPostTime(user *User, timeStr string) (*store
 		return nil, err
 	}
 
+	if user.Settings.DailySummary == nil {
+		user.Settings.DailySummary = &store.DailySummaryUserSettings{
+			PostTime: "8:00AM",
+			Timezone: "Eastern Standard Time",
+			Enable:   false,
+		}
+	}
+
 	dsum := user.Settings.DailySummary
 	dsum.PostTime = timeStr
 	dsum.Timezone = timezone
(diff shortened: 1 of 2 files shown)


server/mscalendar/daily_summary.go
```diff
@@ -75,6 +83,14 @@ func (m *mscalendar) SetDailySummaryEnabled(user *User, enable bool) (*store.Dai
 		return nil, err
 	}
 
+	if user.Settings.DailySummary == nil {
+		user.Settings.DailySummary = &store.DailySummaryUserSettings{
```
jfrerich, line 87: LGTM. Wondering if it would be worth returning this object from a small function since it is used 3 times?

jfrerich (approved):
Non-block suggestion. OTW, LGTM

server/mscalendar/daily_summary.go
larkox: I thought so, but since the object is being used in the store and in the main package, I thought it would create circular dependencies unless we store it in the store. And the store does not seem the best place for this.

mickmister: 1/5 I agree with Jason to make it into a function. If the store is the only place it can go, then I suppose that's where it has to go.

mickmister (changes requested):
LGTM, though I have one request to move the struct initialization into a function

mickmister (approved):

DHaussermann (approved):
Tested and passed
- Tested affected user and saw crash resolved
- Regression tested disconnect and authentication
- No issues found

LGTM!

DHaussermann:
Thanks @larkox for this fix. 

Could you please merge to master so I can pull this into https://github.com/mattermost/mattermost-plugin-mscalendar/pull/183 this issue seems to be complicating testing for the Slack Attachment PR.

PR #209, opened 2021-02-09
Add Autocomplete icon

hanzei:
#### Summary
Screenshot from 2021-02-15 09-49-48
Screenshot from 2021-02-15 09-49-34

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/191


abhijit-singh:
@hanzei The icon seems to have a white background within the image which is inconsistent with the rest of the plugins. Can we replace the icon with the attached file?
ms calendar

hanzei:
@abhijit-singh E.g. `/github` also uses a white background. To me there doesn't seem to be a standard. Still, I'm 0/5 on what to do there. Just wanted to check with you what the standard should me.

abhijit-singh:
Agreed. I can see why it's almost necessary with Github, unless we solve it algorithmically, which I feel is ideal. I'll create a ticket for myself to come up with a logic that works, till then can we take a call on a case-by-case basis?

hanzei:
@abhijit-singh Sure :+1: 

It seems like the file you send above (https://user-images.githubusercontent.com/14369607/107503207-ae2c5a80-6bbf-11eb-9905-a8def040f187.png) also contains a background. Could you please double check?

abhijit-singh:
The file is transparent if you download it and check. It seems like the browser is adding that background colour for some reason.
Screenshot 2021-02-10 at 8 51 18 PM

hanzei:
@abhijit-singh Yes, that seems to be added by my browser. Could you please provide the image as an `scg`? Slash command icon have to be that type.

hanzei:
@abhijit-singh Thanks for the updated image. I've updated the screenshot. Please take another look.

abhijit-singh (approved):

larkox (approved):

iomodo (approved):

DHaussermann (approved):
Tested and passed 
- Autocomplete icon appears as expected
- Confirmed that icon appears as expected and there is no light background visible on dark themes
- Icon appears for all sub-commands

LGTM!

PR #210, opened 2021-02-20
Initial migration to Google Calendar

mickmister:
#### Summary

The main things implemented at the moment are `GetDefaultCalendarView` and `GetMe`. OAuth worked already without much tweaking. I copied the `remote/msgraph` folder to `remote/gcal` and went from there.

I've added blocks like the one below for incremental development, so an error is returned when a remote (gcal) function being used is not implemented yet. Below each of these blocks is the msgraph implementation for reference.

```go
func (c *client) CreateCalendar(remoteUserID string, calIn *remote.Calendar) (*remote.Calendar, error) {
	if true {
		return nil, errors.New("gcal CreateCalendar not implemented")
	}

    // ...msgraph implemention below
```

The goal is to get this foundation in place so we can create HW tickets for Google Calendar. I suggest keeping this migration project on a branch in this repo for now to make it easy to keep things in sync with the core calendar plugin codebase. Maybe it could end up being one repo with build-time flags or something. Getting the ball rolling is important if we want to see this project realized.


hahmadia:
Hoping https://github.com/mattermost/community/discussions/42 can win so this PR can get rolling forward! As you've said `Getting the ball rolling is important if we want to see this project realized.`

server/remote/gcal/call.go
```diff
@@ -0,0 +1,105 @@
...
+	return c.call(method, path, contentType, buf, out)
+}
+
+func (c *client) CallFormPost(method, path string, in url.Values, out interface{}) (responseData []byte, err error) {
+	contentType := "application/x-www-form-urlencoded"
+	buf := strings.NewReader(in.Encode())
+	return c.call(method, path, contentType, buf, out)
+}
+
+func (c *client) call(method, path, contentType string, inBody io.Reader, out interface{}) (responseData []byte, err error) {
```
larkox, line 36: Probably we will never have this situation here, but do we ever consider an empty body? (i.e. nil)

server/remote/gcal/event.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"net/http"
+
+	"github.com/pkg/errors"
+	msgraph "github.com/yaegashi/msgraph.go/v1.0"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+)
+
+func (c *client) GetEvent(remoteUserID, eventID string) (*remote.Event, error) {
```
larkox, line 15: Are these not implemented yet, or is there any api limitation not to implement them?

server/remote/gcal/get_default_calendar_view.go
```diff
@@ -0,0 +1,224 @@
...
+				Response: response,
+			}
+
+			isOrganizer = attendee.Organizer
+		}
+	}
+
+	isAllDay := len(event.Start.Date) > 0 // if Date field is present, it is all-day. as opposed to DateTime field
+
+	return &remote.Event{
```
larkox, line 126: Does it make sense to keep calling that package remote? The way I read it, the remote event is the gcal/mscal event. And I want to convert that to a "mattermost plugin readable event". Right?

server/remote/gcal/handle_webhook.go
```diff
@@ -0,0 +1,87 @@
...
+	Resource                       string `json:"resource,omitempty"`
+	SubscriptionExpirationDateTime string `json:"subscriptionExpirationDateTime,omitempty"`
+	SubscriptionID                 string `json:"subscriptionId"`
+	ResourceData                   struct {
+		DataType string `json:"@odata.type"`
+	} `json:"resourceData"`
+}
+
+func (r *impl) HandleWebhook(w http.ResponseWriter, req *http.Request) []*remote.Notification {
+	// Microsoft graph requires webhook endpoint validation, see
```
larkox, line 30: MSgraph related comment. Not sure if not implemented, or the comment left here.

larkox (approved):
Awesome work Michael. Just a few comments here and there.

server/remote/gcal/event.go
mickmister: @larkox They are just not implemented. I haven't run into anything that seems like a limitation issue on Google's side, but there will probably eventually need to be some changes in the core calendar code to support some feature for both calendars.

server/remote/gcal/get_default_calendar_view.go
mickmister: @larkox Are you talking about `remote.Event` specifically (and possibly other contents of `remote` package), and not the `remote/gcal` or `remote/msgraph` packages? I think it makes sense to have things specifically related to remote communication in those packages. But common data models within `remote` can go elsewhere.

As far as things like `remote.Event`, it is basically based off of Microsoft's data model, so it was originally seen as a "remote" event, but now it is indeed just a "mattermost plugin readable event" as you mentioned. It (and other structs like it) can go into a more shareable package like a new package named `model`.

larkox:
Is the plan to evetually merge gcal into master? I would like that better than creating a new plugin.

mickmister:
> Is the plan to eventually merge gcal into master? I would like that better than creating a new plugin.

@larkox Yes that's the plan. I'd like to add build flags to determine whether the plugin should be compiled for google or microsoft, that way everything can one cohesive implementation in the codebase.

larkox:
Why not compile the plugin for both? As a calendar agregator? So you can link your google and your microsoft account?

mickmister:
@larkox That's definitely feasible. I was just thinking we would market them as separate plugins, but that's not my call 🙂

jfrerich (approved):
This looks like a great start and LGTM!  This was much less to review than it first appeared.  

There are quite a few HW tickets that could come out of this merge!  (for hacktoberfest)

server/remote/gcal/event.go
```diff
@@ -0,0 +1,67 @@
...
+	if true {
+		return nil, errors.New("gcal GetEvent not implemented")
+	}
+
+	e := &remote.Event{}
+
+	err := c.rbuilder.Users().ID(remoteUserID).Events().ID(eventID).Request().JSONRequest(
+		c.ctx, http.MethodGet, "", nil, &e)
+	if err != nil {
+		return nil, errors.Wrap(err, "msgraph GetEvent")
```
levb, line 25: s/msgraph/gcal/

levb (changes requested):
Small change requests from the first cursory review. I am very impressed that almost no changes were required outside `upstream`.

Will now test/debug it with my personal account.

What is the plan on merging to master or branching out?

server/remote/gcal/event.go
mickmister: @levb Notice the beginning block of this function with `if true`, meaning this method is not implemented. The code below the `if true` block is copied directly from the mscalendar implementation, and is only there for reference when the method is to be implemented. This is why it says `msgraph` in this code.

Searching for `if true` in the code reveals 16 spots in my search. So those are basically 16 TODOs on the gcal remote side.

server/mscalendar/daily_summary_test.go
```diff
@@ -148,7 +148,7 @@ func TestProcessAllDailySummary(t *testing.T) {
 				gomock.InOrder(
 					mockPoster.EXPECT().DM("user1_mm_id", "You have no upcoming events.").Return("postID1", nil).Times(1),
 					mockPoster.EXPECT().DM("user2_mm_id", `Times are shown in Pacific Standard Time
-Wednesday February 12
+Wednesday February 12, 2020
```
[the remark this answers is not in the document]
mickmister: > is it a general improvement for both?

Yes, added in this PR as an improvement. It may have also just used it for debugging to know what year an event was. Sorting events was an issue at first and that made it easier to see the exact data.

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/a88921f7bb191854adc2c1f54052eaa9d22919f6/server/mscalendar/views/calendar.go#L30

server/remote/gcal/call.go
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
mickmister: This is not used by the gcal code at all

Issue #221, opened 2021-07-07
Plugin requests Mail.Read and Mail.Send permissions but does not use them


mickmister:
When the user connects their account, the plugin requests the following permissions:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/38dbe19f56f7a6f3cccfbb948a01328d898f32a5/server/remote/msgraph/remote.go#L78-L85

which includes `Mail.Read` and `Mail.Send`. Since the plugin does not require these permissions, they should be removed from this list. `MailboxSettings.Read` should be added as well.

From a discussion on community https://community-daily.mattermost.com/core/pl/syje74dkkpfhuq5khwkb7k7i1r

The task here is remove `Mail.Read` and `Mail.Send` from this array, and add `MailboxSettings.Read` to the array.

PR #222, opened 2021-07-07
Remove mail permissions, add mailbox setting permission

mickmister:
#### Summary

This PR removes the request for `Mail.Read` and `Mail.Send` permissions since the plugin never reads or sends mail. I believe this was a preemptive decision to include these permissions in case we would need to read mail in the future, but currently [yagni](https://martinfowler.com/bliki/Yagni.html).

We haven't documented the need for this permission because the plugin doesn't need it, and I'm not sure how installations have worked without having that granted, since the user connect process should have been blocked by Microsoft.

This also adds the request for `MailboxSettings.Read` since that is required to get the user's timezone information.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/221


larkox (approved):

hanzei:
@mickmister Heads up that you need to fix the tests.

jfrerich (approved):
LGTM 👍

DHaussermann (approved):
This change should not have any impact. The setup should work the same with the existing instruction set. 

This will be covered in `v1.1.0` release testing.  
@mickmister please merge.

PR #223, opened 2021-07-27
[MM-36051] Check for license using SKUs

mickmister:
#### Summary

We now perform the license check in `OnActivate` so the plugin does not start up until the license is valid, or the config is set to `EnableDeveloper`. I've also noted in the README that contributors should set `EnableDeveloper` to enabled.

I also had to move the OAuth settings check into `OnActivate` to make it so the license check is reached. This is because the `OnConfigurationChange` hook is called first, so the OAuth settings error was blocking the `OnActivate` hook from running. I think having both of these checks in `OnActivate` is better than checking for the license on every config change. Maybe we should do the OAuth settings check in both `OnActivate` and `OnConfigurationChange`.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-36051


server/enterprise/license.go
```diff
@@ -0,0 +1,20 @@
+package enterprise
+
+import (
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+const (
+	e20          = "E20"
```
levb, line 9: nit: can these not be imported from a server package?

levb (approved):

server/enterprise/license.go
mickmister: @levb They exist as private constants in the `mattermost-plugin-api`. I don't see them in `mattermost-server`.

I think the right thing to do here (and in Jira plugin https://github.com/mattermost/mattermost-plugin-jira/pull/775) is to actually update the `mattermost-plugin-api` package to include this check here. It's just missing a permutation of helper functions. Should be something like `IsEnterpriseAboveE10OrConfiguredForDevelopment()`, or `isE10ProfessionalEnterpriseOrConfiguredForDevelopment()`. That is what we would be using for any other project that uses the source available license that excludes E10 anyway.

mickmister: https://github.com/mattermost/mattermost-plugin-api/blob/02961f034e242600cdc3ef99a554c06c8b70b3dc/license.go#L67

larkox (approved):

server/enterprise/license.go
mickmister: @levb Do you think it's a big deal to have these hardcoded here, or should we update the helpers in [mattermost-plugin-api](https://github.com/mattermost/mattermost-plugin-api/blob/02961f034e242600cdc3ef99a554c06c8b70b3dc/license.go#L67) to include the functionality we need (to exclude only E10)?

aaronrothschild (approved):

README.md
```diff
@@ -19,7 +19,7 @@
 
 ## License
 
-This repository is licensed under the [Mattermost Source Available License](LICENSE) and requires a valid Enterprise E20 license. See [Mattermost Source Available License](https://docs.mattermost.com/overview/faq.html#mattermost-source-available-license) to learn more.
+This repository is licensed under the [Mattermost Source Available License](LICENSE), and requires a valid Mattermost E20, Professional, or Enterprise license. See our [documentation](https://docs.mattermost.com/about/frequently-asked-questions.html#mattermost-source-available-license) to learn more. If you are contributing to the project, please enable [ServiceSettings.EnableDeveloper](https://docs.mattermost.com/configure/configuration-settings.html#enable-developer-mode) in your server's config. The plugin will not start and show an error messge in your server logs, if you are the missing the Enterprise License.
```
justinegeffen, line 22:
```suggestion
This repository is licensed under the [Mattermost Source Available License](LICENSE), and requires a valid Mattermost E20, Professional, or Enterprise license. See our [documentation](https://docs.mattermost.com/about/frequently-asked-questions.html#mattermost-source-available-license) to learn more. If you are contributing to the project, please enable [ServiceSettings.EnableDeveloper](https://docs.mattermost.com/configure/configuration-settings.html#enable-developer-mode) in your server's config. The plugin will not start and show an error message in your server logs, if you are the missing the Enterprise License.
```

justinegeffen:
Thanks, @mickmister! For the Source Available License wording, could you confirm that it's from Legal and, if so, whether all instances of the content should be updated (if they haven't already). :)

server/plugin/plugin.go
```diff
@@ -64,10 +69,24 @@ func NewWithEnv(env mscalendar.Env) *Plugin {
 }
 
 func (p *Plugin) OnActivate() error {
-	pluginAPIClient := pluginapilicense.NewClient(p.API)
+	pluginAPIClient := pluginapiclient.NewClient(p.API, p.Driver)
+	conf := pluginAPIClient.Configuration.GetConfig()
+	license := pluginAPIClient.System.GetLicense()
+	if !enterprise.HasEnterpriseFeatures(conf, license) {
+		msg := fmt.Sprintf(licenseErrorMessage, config.ApplicationName)
+		return errors.New(msg)
```
hanzei, line 77: Use   `Errorf`?

justinegeffen (approved):
This looks good to me!

server/plugin/plugin.go
mickmister:
```suggestion
		return errors.Errorf(licenseErrorMessage, config.ApplicationName)
```

hanzei (approved):

DHaussermann (approved):
Tested and passed 
- No change in behavior for legacy licenses 
  - Cloud allows the plugin to enable   
  - E20 allows the plugin to enable
  - E10 does **not** allow the plugin to enable
- New SKUs work as expected for Cloud and On-prem
  - Starter does **not** allow the plugin to enable
  - Professional allows the plugin to enable 
  - Enterprise allows the plugin to enable  
 - Logging occurs as expected when plugin fails to start 
 LGTM!

PR #213, opened 2021-03-19
Release bump to v1.1.0

mickmister:
#### Summary

Bump mscalendar plugin to v1.1.0

#### Improvements

b8aff5a Implement autorespond feature (#188)
596cad1 Add Autocomplete icon (#209)
19d0766 Fix autocomplete terms and help text to match available commands (#200)
5798bf3 Add Fallback to Slack Attachments (#183)
b952d3c [GH-125] Remember user last status on status change (#140)
3a6d0e7 Improve message on login flow for receiving notifications during meetings (#158)
61581f3 Use limit 20 for calendar events API call (#178)
5c6f67e Add Telemetry to MSCalendar (#152)
c0d915f Order daily summary view (#177)
3e459f7 Update license to Source Available license (#182)
e05112b Drop "E20 only" from plugin name (#208)

#### Fixes

a488054 Fix panic when Daily Summary is nil for a user (#199)
50beb11 When trying to renew an expired subscription, create new subscription instead (#131)
26fe3c5 use api for license and fix tests (#195)
b8de849 Update system admin check in api.go (#197)
62dad6e Update how to fetch diagnostic id to avoid panics (#187)

#### Maintenance

8437af7 Update README.md (#190)
45c98fd Update plugin.json (#207)
5e9b8e3 Bump lodash from 4.17.11 to 4.17.20 in /webapp (#185)
d10dbd4 Bump elliptic from 6.5.0 to 6.5.3 in /webapp (#186)
47245b0 Bump ini from 1.3.5 to 1.3.8 in /webapp (#201)
4a9daba Issue-184/integrate-golanglint-only (#196)
c751170 Add release_notes_url (#193)
aee00cc go mod tidy (#212)
5cb1e2a Bump elliptic from 6.5.3 to 6.5.4 in /webapp (#211) (showing up twice here)

PR #232, opened 2021-09-22
Add mockgen to go.sum

mickmister:
#### Summary

See https://github.com/mattermost/mattermost-plugin-mscalendar/pull/213#issuecomment-923111661

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1093,6 +1093,7 @@ golang.org/x/mod v0.1.1-0.20191107180719-034126e5016b/go.mod h1:QqPTAvyqsEbceGzB
 golang.org/x/mod v0.2.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.1-0.20200828183125-ce943fd02449/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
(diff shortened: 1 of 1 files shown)


mickmister:
@hanzei The linter is complaining and saying that `go mod tidy` needs to be run. When I run `go mod tidy`, the original error returns:

```
go install github.com/golang/mock/mockgen
../../../../pkg/mod/github.com/golang/mock@v1.4.4/mockgen/parse.go:35:2: missing go.sum entry for module providing package golang.org/x/tools/go/packages (imported by github.com/golang/mock/mockgen); to add:
        go get github.com/golang/mock/mockgen@v1.4.4
make: *** [mock] Error 1
```

I'm not sure what needs to be done here

go.sum
```diff
@@ -1093,6 +1093,7 @@ golang.org/x/mod v0.1.1-0.20191107180719-034126e5016b/go.mod h1:QqPTAvyqsEbceGzB
 golang.org/x/mod v0.2.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.1-0.20200828183125-ce943fd02449/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
+golang.org/x/mod v0.4.2 h1:Gz96sIWK3OalVv/I/qNygP42zyoKp3xptRVCWRFEBvo=
```
hanzei, line 1096: Adding the changes to `go.sum` doesn't seem like the right way to do things. Using a defined version in the `go install` command would ensure the right version is always used and would fix the build error.

hanzei (changes requested):

hanzei (approved):

PR #243, opened 2023-01-12
Fixed lint errors

manojmalik20:
#### Summary
Fixed lint errors


hanzei (changes requested):
Would you please downstream the changes in `.golangci.yml` from https://github.com/mattermost/mattermost-plugin-github/pull/590/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9?

mickmister (approved):

server/jobs/job_manager.go
```diff
@@ -16,26 +16,26 @@ import (
 )
 
 type JobManager struct {
+	env            *mscalendar.Env
+	papi           cluster.JobPluginAPI
 	registeredJobs sync.Map
 	activeJobs     sync.Map
-	env            mscalendar.Env
```
hanzei, line 21: Why was `env` changed to a pointer?

hanzei (approved):
I don't think the field ordering changes for structs are needed given that `fieldalignment` is disabled, but I'm fine with them.

server/jobs/job_manager.go
manojmalik20: The previous `golangci-lint` configuration was giving errors here but now it's not. Do you want me to revert this change?

hanzei: Yes, please

manojmalik20: Done.

Issue #244, opened 2023-01-18
Status sync job exits if an error happens for a single user


mickmister:
Hi, we are having an issue with the mscalender plugin `v1.1.0`. Everything was working fine and then they stopped getting notifications for meetings and status updates. We checked the logs and they were getting the error `"error":"invalid_grant","error_description":"AADSTS700082: The refresh token has expired due to inactivity`. They just generated a new token and installed it, and that resolved the issue with meeting notifications, however status updates are not working (it's not updating their status to indicate they are in a meeting). The error they are getting now in the logs is `Error during user status sync job. err=not found` They have tried restarting the plugin, reconnecting the plugin to their O365 account, and also reconnecting their account.  Any thoughts on what check next?

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/gtyhjtzgmtdb7efqa9cwuzub9w) *by @ewwollesen*._

mickmister:
We should also try to be more descriptive with the error messages from the jobs such as the status sync job. The `not found` error we're currently getting here is not very helpful, though I assume it's coming from here:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/548fd466dd2f00c83df42086ef709f4b4f90a9fc/server/mscalendar/availability.go#L66

Kshitij-Katiyar:
@mickmister By this ``` we would want to limit how many we log```. Here the limit is some kind of number like 100 logs or it means we should restrict (try to reduce) the logs we print.

PR #245, opened 2023-01-30
[GH-244]: Fixed issue "Status sync job exits if an error happens for a single user" by added additional logging

Kshitij-Katiyar:
**Summary** 

- Users were getting the error in the server logs is an Error during user status sync job. err=not found during user status sync job. The status sync job was exiting if an error happened for a single user. So, I changed the logic to continue the process and log the error in server logs.
- There will be at max 5 logs for a particular error/function.

- [Issue #244](https://github.com/mattermost/mattermost-plugin-mscalendar/issues/244)


server/mscalendar/availability.go
```diff
@@ -41,16 +43,17 @@ func (m *mscalendar) Sync(mattermostUserID string) (string, error) {
 func (m *mscalendar) SyncAll() (string, error) {
 	err := m.Filter(withSuperuserClient)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("not able to filter the super user client. %w", err)
```
mickmister, line 46: I think this is a little cleaner. Same with other references of `fmt.Errorf`

```suggestion
		return "", errors.Wrap(err, "not able to filter the super user client")
```

server/mscalendar/availability.go
```diff
@@ -41,16 +43,17 @@ func (m *mscalendar) Sync(mattermostUserID string) (string, error) {
...
 
 	userIndex, err := m.Store.LoadUserIndex()
 	if err != nil {
 		if err.Error() == "not found" {
 			return "No users found in user index", nil
 		}
-		return "", err
+		return "", fmt.Errorf("not able to load users from user index. %w", err)
 	}
+	m.Logger.Debugf("Users loaded successfully from user index")
```
mickmister, line 56: Since this will log on every run of this job, I think we should remove this log here. The improved verbosity of the error messages will work great as a substitute.

server/mscalendar/availability.go
```diff
@@ -60,12 +63,21 @@ func (m *mscalendar) syncUsers(userIndex store.UserIndex) (string, error) {
...
 
+	numberOfLogs := 0
 	users := []*store.User{}
 	for _, u := range userIndex {
 		// TODO fetch users from kvstore in batches, and process in batches instead of all at once
 		user, err := m.Store.LoadUser(u.MattermostUserID)
 		if err != nil {
-			return "", err
+			if numberOfLogs < logTruncateLimit {
+				m.Logger.Errorf("Not able to load user %s from user index. err=%v", u.MattermostUserID, err)
```
mickmister, line 73: There's not a call to action for the admin to fix this, so we should use `Warn` instead

```suggestion
				m.Logger.Warnf("Not able to load user %s from user index. err=%v", u.MattermostUserID, err)
```

server/mscalendar/availability.go
```diff
@@ -60,12 +63,21 @@ func (m *mscalendar) syncUsers(userIndex store.UserIndex) (string, error) {
...
+			} else if numberOfLogs == logTruncateLimit {
+				m.Logger.Warnf(logTruncateMsg)
+			}
+			numberOfLogs++
+
+			// In case of error in loading, skip this user and continue with the next user
+			continue
```
mickmister, line 80: Nice :+1:

server/mscalendar/availability.go
```diff
@@ -157,7 +176,12 @@ func (m *mscalendar) setUserStatuses(users []*store.User, calendarViews []*remot
 			continue
 		}
 		if view.Error != nil {
-			m.Logger.Warnf("Error getting availability for %s. err=%s", user.Remote.Mail, view.Error.Message)
+			if numberOfLogs < logTruncateLimit {
+				m.Logger.Warnf("Error getting availability for %s. err=%s", user.Remote.Mail, view.Error.Message)
```
mickmister, line 180: Unrelated to the main purpose of this PR, but let's change this to the Mattermost user ID instead of the user's email. We shouldn't be logging this personal information, especially since it's coming from an external system.

Same with other occurrence of logging this value

server/mscalendar/availability_test.go
```diff
@@ -118,6 +118,7 @@ func TestSyncStatusAll(t *testing.T) {
 
 			c, papi, s, logger := client.(*mock_remote.MockClient), deps.PluginAPI.(*mock_plugin_api.MockPluginAPI), deps.Store.(*mock_store.MockStore), deps.Logger.(*mock_bot.MockLogger)
 
+			logger.EXPECT().Debugf("Users loaded successfully from user index")
```
mickmister, line 121: The changes in this file will be reverted when the debug log is removed

mickmister (changes requested):
@Kshitij-Katiyar Thanks for the quick turnaround on this :+1: Mostly looks good, just a few requests

mickmister (approved):
Awesome thanks @Kshitij-Katiyar!

DHaussermann:
@Kshitij-Katiyar one suggestion to consider here would be - Once the sync job completes, can it write to the logs how many users where sync and how many errored out? This would provide visibility and confirm that the the job is now attempting to sync the entire user base.

Kshitij-Katiyar:
> @Kshitij-Katiyar one suggestion to consider here would be - Once the sync job completes, can it write to the logs how many users where sync and how many errored out? This would provide visibility and confirm that the the job is now attempting to sync the entire user base.

sure @DHaussermann We can do that after the job ends, we can add a summary for that job. Can you please also give suggestions of what are the things I should include in that? Some examples could be
1) Can only show the number of users for each category like status synced-10 users and error-5 users.
2) Along with the number we can show the mattermost userId for each user. I don't think it would be better to show usernames as during the process we only have user Ids during this process and for each user, I have to call an additional API to fetch the usernames.

Please give your views on this.

mickmister:
@Kshitij-Katiyar @DHaussermann I think we want to be cognizant of logging too much here. At the moment, we have a debug log before the job runs, and one after the job runs. These logs simply say that the job began and ended. There are some different levels of granularity we can add here:

- Log how many users processed, and how many had errors
- Include how many users had their status changed

The individual errors have the user ids in them, so no need to include the user ids in the aggregate log message at the end of the job. I think the first bullet point is appropriate here. If the second one is not a big hassle, then I think we should also include the "number of statuses changed".

Kshitij-Katiyar:
@mickmister @DHaussermann Added the functionality of showing the summary for the status sync job

mickmister (approved):

server/config/config.go
```diff
@@ -8,25 +8,22 @@ type StoredConfig struct {
...
 	BuildDate              string
 	BuildHash              string
 	BuildHashShort         string
 	MattermostSiteHostname string
 	MattermostSiteURL      string
-	PluginID               string
 	PluginURL              string
 	PluginURLPath          string
 	PluginVersion          string
+	StoredConfig
```
hanzei, line 28: Unrelated change?

server/mscalendar/availability.go
```diff
@@ -21,78 +22,101 @@ const (
...
 	for _, u := range userIndex {
 		// TODO fetch users from kvstore in batches, and process in batches instead of all at once
 		user, err := m.Store.LoadUser(u.MattermostUserID)
 		if err != nil {
-			return "", err
+			if numberOfLogs < logTruncateLimit {
+				m.Logger.Warnf("Not able to load user %s from user index. err=%v", u.MattermostUserID, err)
+			} else if numberOfLogs == logTruncateLimit {
+				m.Logger.Warnf(logTruncateMsg)
+			}
```
hanzei, line 84: What is the value of posing a truncated message instead of the "real" one?

hanzei (changes requested):
There are a lot of unrelated changes in the PR. Could you please remove them?

server/config/config.go
Kshitij-Katiyar: @hanzei The branch was not in sync with the master.Now it is fixed

server/mscalendar/availability.go
Kshitij-Katiyar: @hanzei Please read this comment https://github.com/mattermost/mattermost-plugin-mscalendar/issues/244#issuecomment-1402281372.

hanzei: @mickmister I don't understand your point in https://github.com/mattermost/mattermost-plugin-mscalendar/issues/244#issuecomment-1402281372. If the goal is not to flood the logs, why are messages logged that only contain the information that the logs got trunkated? Isn't the number of log messages still the same?

Kshitij-Katiyar: @hanzei No the number of logs will not be the same. We will log the first 4 logs coming from a particular function/code snippet, then for the 5th one we will log the Truncate and after that, we will not be logging anything. So the max number of logs will be 5 in any case. If it is more than that we are truncating means not logging anything for a particular function.

Kshitij-Katiyar:
> @Kshitij-Katiyar Heads up that there is a linter issue

@hanzei Fixed lint errors.

Kshitij-Katiyar:
@DHaussermann @mickmister  I have fixed this. Actually in the code at one point I was accessing the attributes of a nil object, resulting in panic. This could only happen in an instance where there was no previous ms-calendar plugin installed. I have added a check now and also tested it on a new instance everything worked fine. Please test this again and let me know if you find any issues.

DHaussermann:
Hi @Kshitij-Katiyar 
I did some testing on this today. 
- The panic issue is fixed for new installs
- I can see that the sync service starts and logs an output when it runs

The piece I'm struggling to test here is to see and error occur on 1 of the records and the job will still complete. from your summary "_`err=not found during user status sync job..._"  How can I reproduce such an error?

I have this running locally with 2 users and have tried manually updating several values in KV store to invalidate them to try and cause such an error so I can see the sync complete. So far this has not worked. Updating `oauth2_` values or `user_` to try and invalidate them does cause the user to loose their connection but it still does not cause the job to error out on any of the records. 

Are you able to offer any suggestions here?

Kshitij-Katiyar:
> Hi @Kshitij-Katiyar I did some testing on this today.
> 
> * The panic issue is fixed for new installs
> * I can see that the sync service starts and logs an output when it runs
> 
> The piece I'm struggling to test here is to see and error occur on 1 of the records and the job will still complete. from your summary "_`err=not found during user status sync job..._" How can I reproduce such an error?
> 
> I have this running locally with 2 users and have tried manually updating several values in KV store to invalidate them to try and cause such an error so I can see the sync complete. So far this has not worked. Updating `oauth2_` values or `user_` to try and invalidate them does cause the user to loose their connection but it still does not cause the job to error out on any of the records.
> 
> Are you able to offer any suggestions here?

Hi @DHaussermann , I have done the changes for the part that even if we get an error from 1 user the job will still continue for other users. If you want to get or reproduce the above error **user not found** I can tell you I tested it. Steps for repro - 
1) Connect the user using the slash command.
2) You can delete the key value for that user from the KV store either by going inside DB (if you can) and run ```DELETE FROM pluginkeyvaluestore where pluginid ='com.mattermost.mscalendar' AND pkey = 'user_<key for your user>';
```  or using MM APIs. 

![Screenshot from 2023-03-21 13-05-07](https://user-images.githubusercontent.com/90389917/226543717-c27c5004-3720-4a32-a0a3-816272aeeb59.png)

```
{"timestamp":"2023-03-21 13:10:01.485 +05:30","level":"warn","msg":"Not able to load user 3p41xahsffbh9kqk1jcx9j1tpw from user index. err=not found","caller":"app/plugin_api.go:943","plugin_id":"com.mattermost.mscalendar"}
```

**Comment by Kshitij-Katiyar** [comment]
> Thanks @Kshitij-Katiyar taking another look and using your suggestion of deleting the records from the store - I can see the user not found errors
> 
> ```
> warn  [2023-03-21 10:10:00.389 -04:00] Not able to load user 6sbsdz3rgprs3f8s64nxft8s5a from user index. err=not found caller="app/plugin_api.go:979" plugin_id=com.mattermost.mscalendar
> warn  [2023-03-21 10:10:00.399 -04:00] Not able to load user q6cs674637nguebthxp5r8m1rc from user index. err=not found caller="app/plugin_api.go:979" plugin_id=com.mattermost.mscalendar
> debug [2023-03-21 10:10:00.399 -04:00] User status sync job finished.
> Summary
> Number of users processed:- 2
> Number of users had their status changed:- 0
> Number of users had errors:- 2 caller="app/plugin_api.go:970" plugin_id=com.mattermost.mscalendar
> ```
> 
> The issue yesterday was that `Number of users had errors:-` would continue showing as `0`. I'm now not certain if `Not able to load user <user ID> from user index` was happening or not but even if it was, it would always show the error count as 0. Maybe your latest change may have affected this.
> 
> Also @mickmister does this last commit require code review?

@DHaussermann Ok yes I have fixed it. It was a very small change.

File: server/mscalendar/availability.go [hunk 9]
```diff
@@ -39,33 +39,34 @@ type Availability interface {
...
 		if err.Error() == "not found" {
-			return "No users found in user index", nil, nil
+			return "No users found in user index", syncJobSummary, nil
 		}
-		return "", nil, errors.Wrap(err, "not able to load the users from user index")
+		return "", syncJobSummary, errors.Wrap(err, "not able to load the users from user index")
 	}
 
-	return m.syncUsers(userIndex)
+	return m.syncUsers(userIndex, syncJobSummary)
```
hanzei, line 66: syncJobSummary is always an empty value. Why is passed here?

Kshitij-Katiyar: @hanzei Passed as a reference in the syncUsers function as I am updating the values of syncJobSummary in syncUsers function.

Kshitij-Katiyar: but yeah I can remove it, so removed it.

server/mscalendar/availability.go
```diff
@@ -21,78 +22,102 @@ const (
...
 
 	return m.syncUsers(store.UserIndex{user})
 }
 
-func (m *mscalendar) SyncAll() (string, error) {
+func (m *mscalendar) SyncAll() (string, *StatusSyncJobSummary, error) {
 	err := m.Filter(withSuperuserClient)
 	if err != nil {
-		return "", err
+		return "", &StatusSyncJobSummary{}, errors.Wrap(err, "not able to filter the super user client")
```
hanzei, line 53: Nit pick: In general, functions that return a non-nil error should have the other fields set to `nil`.

hanzei (approved):

server/mscalendar/availability.go
Kshitij-Katiyar: @hanzei  Actually here I am not returning nil because, in its parent functions, I am using an attribute of this object, and if I return nil from here that could give panic.

DHaussermann:
@mickmister or @hanzei please merge when you have a chance. 
We also need a v`1.1.1` versions bump PR.

Kshitij-Katiyar:
> /update-branch

@mickmister @hanzei Synced with master

PR #253, opened 2023-03-31
telemetry reload + migration to server/v6

trilopin:
#### Summary
Fix telemetry problem with hooks race condition between Onactivate and OnConfigurationChange

To use new pluginapi I had to update from server/v5 to server/v6 and ended up doing more changes than I wanted to.
Other relevant changes:
- compile for arm64 archs
- migrated local telemetry package to use same approach as jira
- `make server` now has the mechanism to handle current-arch and mmdebug 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-51818


server/plugin/plugin.go
```diff
@@ -92,11 +93,30 @@ func (p *Plugin) OnActivate() error {
...
+				p.API.GetServerVersion(),
+				e.PluginID,
+				e.PluginVersion,
+				config.TelemetryShortName,
+				telemetry.NewTrackerConfig(p.API.GetConfig()),
+				logger.New(p.API),
+			),
+		)
+		e.bot = e.bot.WithConfig(stored.Config)
+		e.Dependencies.Store = store.NewPluginStore(p.API, e.bot, e.Dependencies.Tracker)
```
trilopin, line 117: I've had to add this to prevent a panic from happening.

Sometimes the hooks are called: onConfChange + Onactivate (without a second onConfChange), and that ended up with nil trackers stored across subpackages like flow.

server/utils/bot/bot.go
```diff
@@ -54,8 +53,8 @@ func (bot *bot) Ensure(stored *model.Bot, iconPath string) error {
 		// Already done
 		return nil
 	}
-
-	botUserID, err := bot.pluginHelpers.EnsureBot(stored, plugin.ProfileImagePath(iconPath))
+	client := pluginapi.NewClient(bot.pluginAPI, nil) // driver passed as nil, as we don't need it
```
mickmister, line 56: > as we don't need it

Could this assumption change in the future? I'm actually not sure what the `driver` is used for.

mickmister (approved):
Great work @trilopin :tada:

server/utils/bot/bot.go
trilopin: That driver is used for doing direct database queries, and the this created client is used just to do EnsureBot (which don't use direct). 

But still I'm 2/5 too... What's worse? to cross a dep to driver (which we don't need) or to bet on nil-driver hoping nobody uses it in the future...

trilopin:
@DHaussermann I don't usually go through QA review in telemetry PRs. However, I found (and fixed) some panics in this one + the v6 migration..I think it'd be nice to have your brain on this.

trilopin:
@hanzei kind reminder for review 🙏🏽

DHaussermann (approved):
Tested and passed
- Deployed and smoke tested the plugin
  - Connect / Disconnect
  - Summary and sync jobs run
  - Using slash commands to retrieve date from MS Calendar side is working
- No panics or other crashes found
- No functional issues found
- Confirmed `make` now also compiles a `darwin-arm64` 

LGTM!

**note** @trilopin I did not test telemetry for this PR. We can follow up if this is needed as part of release testing.

hanzei (commented):
If we are updating to a newer server version, should we also bump the `min_server_version`? Was the plugin tested against 5.37.0?

plugin.json
```diff
@@ -7,11 +7,13 @@
     "release_notes_url": "https://github.com/mattermost/mattermost-plugin-mscalendar/releases/tag/v1.2.1",
     "icon_path": "assets/profile.svg",
     "version": "1.2.1",
-    "min_server_version": "5.37.0",
+    "min_server_version": "6.3.0",
```
trilopin, line 10: note this change has been done

trilopin:
@mickmister asked again for review, min_server_version updated to 6.3, and telemetry copied again inside the repo until we can upgrade to 7x(pluginapi 0.1.3).

mickmister (approved):

hanzei (approved):

Issue #198, opened 2020-11-10
Refresh token expiring due to user inactivity


mickmister:
Per the discussion in https://github.com/mattermost/mattermost-plugin-mscalendar/issues/15#issuecomment-585033566, it was determined that we should allow the OAuth2 library to handle access tokens and refresh tokens with the library's default behavior.

The refresh token is used to retrieve an access token whenever we need to perform an operation in Microsoft's system. A customer's log reports an error that says ` The refresh token has expired due to inactivity`. This error is occurring during the Renew Subscription job. The customer said this error is printed to the logs every night at midnight.

> (log ERROR) Error renewing subscription. err=msgraph RenewSubscription: Patch "https://graph.microsoft.com/v1.0/subscriptions/REDACTED": oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description":"AADSTS700082: The refresh token has expired due to inactivity. The token was issued on 2020-07-23T12:22:48.1096407Z and was inactive for 90.00:00:00.\r\nTrace ID: b40f4014-73bb-40f3-b90d-cd42d78c9600\r\nCorrelation ID: 1bb86522-a8e8-45a2-9966-2d11a892c925\r\nTimestamp: 2020-11-05 00:00:00Z","error_codes":[700082],"timestamp":"2020-11-05 00:00:00Z","trace_id":"b40f4014-73bb-40f3-b90d-cd42d78c9600","correlation_id":"1bb86522-a8e8-45a2-9966-2d11a892c925","error_uri":"https://login.microsoftonline.com/error?code=700082"}

Microsoft's [docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-configurable-token-lifetimes#token-lifetimes-with-confidential-client-refresh-tokens) state that their refresh tokens do not expire. Here are two routes we may take:

- Auto-renew refresh tokens. I think this may be undesirable on Microsoft's side. But other than that I don't think it's a problem.
- Handle this error when it happens, mark the user as inactive locally, and have the user re-authenticate the next time they try to do something with the plugin.

More context here https://community-daily.mattermost.com/core/pl/u7jyh7dokpdtfcq49keu6peofe

_Issue created from a [Mattermost message](https://community-daily.mattermost.com/_redirect/pl/iwcsqwh8j38cigj1zasqhn9nkr) *by @mickmister*._

coltoneshaw:
[...]

PR #256, opened 2023-04-19
[GH-198]:Fixed issue #198 'Handle case where a user's refresh token expires'

Kshitij-Katiyar:
**Summary** 

- Added the logic that if an API request fails, with the error refresh token expired then marking the user inactive in the KV store and also sending a DM to the user to reconnect their account.

- If the user is marked as inactive, aborting any future API requests done for this user, and logging via the WARN log level.

- [Issue #198](https://github.com/mattermost/mattermost-plugin-mscalendar/issues/198)


Kshitij-Katiyar:
@m1lt0n Not able to add you as a reviewer here.

m1lt0n (approved):

server/remote/msgraph/get_notification_data.go
```diff
@@ -9,17 +9,24 @@ import (
...
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
 )
 
 func (c *client) GetNotificationData(orig *remote.Notification) (*remote.Notification, error) {
 	n := *orig
 	wh := n.Webhook.(*webhook)
+	if !c.CheckUserStatus() {
+		c.Logger.Warnf(LogUserInactive, c.mattermostUserID)
+		return nil, errors.New(ErrorUserInactive)
+	}
```
mickmister, line 22: I think we would be sending the user a DM for every incoming notification here, telling them that they need to reconnect

server/remote/msgraph/get_me.go
```diff
@@ -37,3 +45,50 @@ func (c *client) GetMe() (*remote.User, error) {
...
+
+	// Checking if the user is marked as inactive
+	if user.OAuth2Token.AccessToken == "" {
+		if _, err := c.poster.DM(c.mattermostUserID, ErrorUserInactive); err != nil {
+			c.Logger.Errorf("Not able to DM the user %s. %s", c.mattermostUserID, err.Error())
+		}
+		return false
+	}
+
+	return true
```
mickmister, line 69: This and `ChangeUserStatus` seems wrong to be in a file called `server/remote/msgraph/get_me.go`. I think the logic pertaining to the store etc. should be in a different package.

mickmister (changes requested):
Great work on this @Kshitij-Katiyar :+1:

I have a few requests mainly pertaining to avoiding DM'ing a user multiple times, and putting in some defensive code to avoid token expiry issues.

server/remote/msgraph/get_me.go
```diff
@@ -4,12 +4,20 @@
...
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/serializer"
+
 	"github.com/pkg/errors"
+)
 
-	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+const (
+	ErrorUserInactive        = "You have been marked inactive. Please disconnect and reconnect your account again."
+	LogUserInactive          = "User %s is inactive. Please disconnect and reconnect your account."
+	ErrorRefreshTokenExpired = "The refresh token has expired due to inactivity"
```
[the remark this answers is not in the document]
Kshitij-Katiyar: @mickmister Regarding this. Usually, the flow is we get a refresh and access token from the Oauth. If the access token is expired we can call the API and get the updated access and refresh token (Usually get handled by the Oauth golang library). But if both the access and refresh token is expired we have to reconnect the account again. 

I don't understand what I need to do in this case. Here auto-renewal means that should I implement a job that will run after some set interval that will update the refresh token for all the connected users or if there is something else I needed to do to automate this.

Please explain this a bit.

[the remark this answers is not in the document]
mickmister: > Usually, the flow is we get a refresh and access token from the Oauth. If the access token is expired we can call the API and get the updated access and refresh token (Usually get handled by the Oauth golang library)

@Kshitij-Katiyar Yes the token refresh is handled by the oauth library, but our code needs to _store_ the refresh token when we receive a new one during the access token refresh. We should be updating the KV store with new refresh tokens, every time we get one from msgraph.

server/remote/msgraph/get_me.go
```diff
@@ -4,15 +4,13 @@
...
-	"strings"
-
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/serializer"
 
 	"github.com/pkg/errors"
 )
 
 const (
-	ErrorUserInactive        = "You have been marked inactive. Please disconnect and reconnect your account again."
+	ErrorUserInactive        = "You have been marked inactive because your refresh token is expired. Please disconnect and reconnect your account again."
```
mickmister, line 13: This actually reads kind of weird, if this is an error we show to the user. I think we can leave that part out of this string. We can log a different string for this. Also, I think we should move these strings to a different file since they're not used in this file.

server/remote/msgraph/check_user_status.go
```diff
@@ -0,0 +1,55 @@
...
+
+import "strings"
+
+func (c *client) CheckUserStatus() bool {
+	// Not checking for API calls made using super client
+	if c.store == nil {
+		return true
+	}
+
+	user, err := c.store.LoadUser(c.mattermostUserID)
```
mickmister, line 11: I still think this code belongs in a different package than `remote/msgraph`. This code doesn't have much to do with `msgraph`, compared to the business logic of the store etc. This package should not be aware of the store.

mickmister:
@Kshitij-Katiyar This issue has become a more pressing issue, so I would prioritize this PR if possible. There some remaining change requests, though the main one is here https://github.com/mattermost/mattermost-plugin-mscalendar/pull/256#discussion_r1178289407

server/remote/msgraph/get_me.go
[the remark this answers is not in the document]
mickmister: Here's an example from the Zoom plugin storing the refresh token:
https://github.com/mattermost/mattermost-plugin-zoom/blob/8ab42cf6f19a93f463065a57af874eb36bf2b6cd/server/zoom/oauth.go#L131-L147

```go
currentToken := info.OAuthToken

tokenSource := c.config.TokenSource(context.Background(), currentToken)
updatedToken, err := tokenSource.Token()
if err != nil {
    return nil, errors.Wrap(err, "error getting token from token source")
}

if updatedToken.RefreshToken != currentToken.RefreshToken {
    info.OAuthToken = updatedToken
    kvErr := c.api.UpdateZoomOAuthUserInfo(user.Id, info)
    if kvErr != nil {
        return nil, errors.Wrap(kvErr, "error setting new token")
    }
}

c.token = updatedToken
```

---

In the case of the mscalendar plugin, I'm thinking we can provide an `onRefresh` parameter to the `MakeClient` function, and store the function on the `client` struct:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/847134e1ff36837676779abc97d9a0a8e79f94c3/server/remote/msgraph/remote.go#L39

```go
func (r *impl) MakeClient(ctx context.Context, token *oauth2.Token, onRefresh func(token *oauth2.Token)) remote.Client {
    c := &client{
		conf:       r.conf,
		ctx:        ctx,
		httpClient: httpClient,
		Logger:     r.logger,
		rbuilder:   msgraph.NewClient(httpClient),
		onRefresh:  onRefresh,
	}
	return c
```

Then we can extract the token storage logic to a different package of the program.

server/remote/msgraph/check_user_status.go
mickmister: See https://github.com/mattermost/mattermost-plugin-mscalendar/pull/256#discussion_r1190320663 on a solution for this

mickmister (approved):

mickmister (approved):

mickmister:
@Kshitij-Katiyar Can you please provide testing steps for this PR for @DHaussermann

Kshitij-Katiyar:
> 



> @Kshitij-Katiyar Can you please provide testing steps for this PR for @DHaussermann

@DHaussermann Testing steps:- 
- First connect the user using command ```/mscalender connect```. 
- Wait for the refresh token to expire. I have manually deleted that that's why I don't know the time for that but I can search for it.
- If the user's token is expired, try to run any slash command which will call the ms-calender API like CRUD operations of the calendar, etc.
- Then the user should get the DM and in the server log stating that the user is marked inactive, please disconnect and connect the account again.
- After this, all APIs will start working again.

DHaussermann:
Hi @Kshitij-Katiyar 
Yes I was curious about the step of waiting for the refresh token to expire. I see the token value in the DB store but I think the expiry must occur on the MS side correct?

Also I'm a bit stuck here because the token could take possibly weeks to expire and simply by invalidating it in the data store, am I thang just hitting a different code path?

If so, I'm not sure if there is functional testing I can cover here. Maybe some type of test build that will mimic the refresh process is an option?

DHaussermann:
Heads-up @Kshitij-Katiyar, neither of the people you mentioned are available. @jasonblais may be able to suggest a resource who can help move this forward. If not, @mickmister will return from PTO but that is a least a couple weeks away.

jasonblais:
Thanks both, is this something that can wait 2 weeks for when Michael is back & can provide feedback? @Kshitij-Katiyar @DHaussermann

DHaussermann:
@jasonblais sorry for the delay.

No, I don't know that this is urgent. The customer issue that was being discussed in the channel recently was solved last month and is not related to this. 
I'm sure there is some customer impact (possibly they need to re-connect their accounts every 2 months) but I see no discussions about it now.

This is an issue we've know about for a long time.

mickmister:
@DHaussermann Note that there are two issues the customer was experiencing:

- Status change and daily summary - Issue caused by service account policy
- Event invitations - Issue caused by expired OAuth token

We've confirmed that the first bullet is solved on their end by changing some Azure environment policy. AFAIK the second bullet is not confirmed, and I think it requires this PR to fix it. There's an active thread discussing this here https://community.mattermost.com/core/pl/xo5k64m6ctremdzdsxbecesdfo. It sounds like it might be fixed but I'm not sure yet

mickmister:
> It sounds like it might be fixed but I'm not sure yet

Let me clarify, I'm pretty certain this is _not_ fixed, but still verifying

Kshitij-Katiyar:
> > Hi @Kshitij-Katiyar Yes I was curious about the step of waiting for the refresh token to expire. I see the token value in the DB store but I think the expiry must occur on the MS side correct?
> > Also I'm a bit stuck here because the token could take possibly weeks to expire and simply by invalidating it in the data store, am I thang just hitting a different code path?
> > If so, I'm not sure if there is functional testing I can cover here. Maybe some type of test build that will mimic the refresh process is an option?
> 
> @DHaussermann I explored a bit on this. The refresh token expires in 60 Days and I also tried to find out if we can configure its expiry time on the Azure side but it's clearly written that it is not possible.
> 
> In this PR I am checking the error coming from the Azure side that the refresh token is expired or not. What we can do is we can invalidate the refresh token from DB and add the condition for checking the error for an invalid token this would simulate the same things as this PR. I can build a test build for this code. Please give your views on this @mickmister @m1lt0n

@mickmister Can u please take a look into this comment and give your thoughts on this.

mickmister:
@Kshitij-Katiyar  Let's try the test build you're referring to. If you think we can have a decent setup that simulates the refresh token expiring, we can have a test build to test this functionality. Otherwise, we may not be able to test this piece and will need to move forward without testing it. cc @DHaussermann

Kshitij-Katiyar:
> @Kshitij-Katiyar Let's try the test build you're referring to. If you think we can have a decent setup that simulates the refresh token expiring, we can have a test build to test this functionality. Otherwise, we may not be able to test this piece and will need to move forward without testing it. cc @DHaussermann

@mickmister @DHaussermann I have created a test build and sent to both of you thorough DM on community.mattermost.com as here the limit for attachements was only 25MB. Please test it and let me know if you face any problem

mickmister:
@Kshitij-Katiyar @DHaussermann Just checking to see what next steps are here. It seems impossible for us to truly test this, as I believe we would need a valid token that is genuinely 60 days old, in order to have msgraph invalidate the token and issue a new refresh token.

@DHaussermann I'm wondering what's needed to get this merged. Not sure what's left to test other than general feature/regression testing. Anything @Kshitij-Katiyar can do or clear up on this?

Kshitij-Katiyar:
> @Kshitij-Katiyar @DHaussermann Just checking to see what next steps are here. It seems impossible for us to truly test this, as I believe we would need a valid token that is genuinely 60 days old, in order to have msgraph invalidate the token and issue a new refresh token.
>  
> @DHaussermann I'm wondering what's needed to get this merged. Not sure what's left to test other than general feature/regression testing. Anything @Kshitij-Katiyar can do or clear up on this?

> @mickmister I don't think there is anything we can do to test this PR specifically. I agree with you we should go with the regression testing.

hanzei:
@Kshitij-Katiyar Would you please merge `master` into your PR?

hanzei:
@DHaussermann Gentle ping

Kshitij-Katiyar:
> @DHaussermann Gentle ping

@mickmister @DHaussermann @hanzei should Brightscout handle the QA part for this PR?

mickmister:
@Kshitij-Katiyar Yes that sounds good :+1:

raghavaggarwal2308:
@mickmister This PR contains changes that QA cannot test directly. We will need to make changes in the code to test it (Like manually decreasing the expiry time or deleting the token from the KV store), which has already been done while development. What are your opinions on skipping the QA review on this PR?

mickmister:
@raghavaggarwal2308 Are we able to write a test that ensures this works correctly?

Kshitij-Katiyar:
> @raghavaggarwal2308 Are we able to write a test that ensures this works correctly?

@mickmister Writing the test cases for this will be a little difficult as first of all there are no existing test cases present for the ms graph package and secondly, we are using an SDK package for making API calls to the external portal and we have to mock it to return a particular refresh token expired error. To do that we have to create a wrapper function above all the API calls and include those in an interface and only then we can create mocks for it secondly we can use monkey patching but that is not that stable.

calendar/store/user_store.go
```diff
@@ -295,6 +305,79 @@ func (s *pluginStore) StoreUserActiveEvents(mattermostUserID string, events []st
 	return kvstore.StoreJSON(s.userKV, mattermostUserID, u)
 }
 
+// RefreshAndStoreToken checks whether the current access token is expired or not. If it is,
+// then it refreshes the token and stores the new pair of access and refresh tokens in kv store.
+func (s *pluginStore) RefreshAndStoreToken(token *oauth2.Token, oconf *oauth2.Config, mattermostUserID string) (*oauth2.Token, error) {
```
mickmister, line 310: @Kshitij-Katiyar This is where much of the functionality was added in this PR. I'm thinking we can test this function and `CheckUserConnected` directly. What do you think?

We basically want to emulate this in a test since this is the code that is applied throughout the PR:

```go
	if !c.tokenHelpers.CheckUserConnected(c.mattermostUserID) {
		c.Logger.Warnf(LogUserInactive, c.mattermostUserID)
		return nil, errors.New(ErrorUserInactive)
	}
	err := someCallToAPI // this can be whatever we want in the test
	if err != nil {
		c.tokenHelpers.DisconnectUserFromStoreIfNecessary(err, c.mattermostUserID)
```

We can use a "real" KV store in the test with one of the following strategies:
- https://gist.github.com/mickmister/f2ac1c99c63291692b18fdae715338af
- https://github.com/mattermost/mattermost-plugin-github/pull/745

Kshitij-Katiyar: @wiggin77 Writing test cases for these functions without refactoring is challenging because many of the functions used within them are standalone and not implemented by any interface, making it difficult to mock them.

One workaround is to use monkeyPatch. However, this approach isn't currently used in the plugin, and introducing it just for this test case might not be ideal. Additionally, no test cases have been added for any `store` or `msgraph` functions so far, possibly because adding them is not straightforward.

wiggin77: This method can be unit tested with a memory store.

Kshitij-Katiyar: Hi @wiggin77,
I looked into the memoryStore and its implementation in Mattermost. It seems useful for mocking the config store in Mattermost and other similar stores within Mattermost plugins. However, I’m unsure how it applies to our current use case. Could you provide some guidance on this?

wiggin77: A solution needs to be found here.  If the implementation does not lend itself to writing clean unit tests, then the implementation needs to be refactored.

Kshitij-Katiyar:
There was an issue with the execution slash commands involving any external API call if executed with an expired or invalid refresh token. Fixed it in this [commit](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/256/commits/fa93af6913d826e71031282fbd20ee00b500a6f2) by checking for the expired token error in execution of related slash commands

raghavaggarwal2308 (approved):

raghavaggarwal2308:
@wiggin77 This PR has been dev tested properly. I do not think we need to get it QA tested as it won't be readily tested because we have to manually go into the DB and change the refresh token multiple times. Please let me know your thoughts on merging this?

wiggin77:
> Please let me know your thoughts on merging this?

Ok

PR #268, opened 2023-07-25
[GCAL] Build flags

fmartingr:
- Restores plugin.json and the profile images to their originals
- Adds a separate plugin-gcal.json manifest for the google calendar plugin
- Allows the use of a `PROVIDER=gcal` environment variable to call the `Makefile` to build the proper plugin each time
- [ ] `build/bin/manifest id` retrieves `com.mattermost.mscalendar` because the `shell` make command does not send environment variables (apparenly) (check `custom.mk`)
- Added the `Provider` build flag to know internally which flavor of the plugin should load.


assets/profile.png
mickmister, line 1: 0/5 Maybe name this `profile-mscalendar.png` for differentiation

build/custom.mk
```diff
@@ -14,6 +14,7 @@ BUILD_HASH_SHORT = $(shell git rev-parse --short HEAD)
 LDFLAGS += -X "main.BuildDate=$(BUILD_DATE)"
 LDFLAGS += -X "main.BuildHash=$(BUILD_HASH)"
 LDFLAGS += -X "main.BuildHashShort=$(BUILD_HASH_SHORT)"
+LDFLAGS += -X "main.Provider=$(PROVIDER)"
```
mickmister, line 17: I wonder if we want to be more specific with `CALENDAR_PROVIDER` instead. There isn't another concept of "Provider" in the plugin project, though probably better for a new dev when reading the code

plugin-gcal.json
```diff
@@ -0,0 +1,102 @@
+{
+    "id": "com.mattermost.gcal",
+    "name": "Google Calendar",
+    "description": "Google Calendar Integration",
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-mscalendar",
+    "support_url": "https://github.com/mattermost/mattermost-plugin-mscalendar/issues",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-mscalendar/releases/tag/v1.0.0",
+    "icon_path": "assets/profile-gcal.svg",
+    "version": "1.0.0",
```
mickmister, line 9: I was thinking about versioning, and I think we should have the two plugin builds follow the same versioning. That way we can track what master changes they both have. 0/5 on this atm

server/config/config.go
```diff
@@ -18,6 +18,7 @@ type StoredConfig struct {
 // Config represents the the metadata handed to all request runners (command,
 // http).
 type Config struct {
+	Provider               string
```
mickmister, line 21: Same here, I think `CalendarProviderName` would be a good name for this. `Name` may be overkill though. I think we'll want some sort of struct like `CalendarProviderMetadata` to store bot username info etc.

Makefile
```diff
@@ -121,7 +127,7 @@ else
 endif
 	rm -rf dist/
 	mkdir -p dist/$(PLUGIN_ID)/server/dist
-	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
+	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/plugin.json
```
mickmister, line 130: LGTM :+1: Nice work

mickmister (commented):
This is looking aweomse :tada: A few comments from a first pass of this

plugin-gcal.json
fmartingr: Unsure about this. My gut tells me that we should do individual versioning, and two releases when the logic is common to both, but also 0/5 on this.

providers/GoogleCalendar.md
```diff
@@ -0,0 +1,34 @@
...
+## Features
+
+- Receive a daily summary at a specific time
+- Receive event reminders 5 minutes before a meeting via direct message
+- Receive direct message notification on event invitations
+- Automatically set an user status (away, DND) during meetings
+
+## Configuration
+
+1. Create a project in the Google Cloud Cnosole
```
mickmister, line 20:
```suggestion
1. Create a project in the Google Cloud Console
```

providers/GoogleCalendar.md
```diff
@@ -0,0 +1,34 @@
...
+    - Go to [console.cloud.google.com](https://console.cloud.google.com/) and click on the dropdown at the top of the page to create a new project.
+2. When you have your project ready, the required APIs need to be enabled on it:
+    - Go to **APIs & Services** search and enable two services:
+        - **Google Calendar API**: Used for anything related to the calendar and events
+        - **Google People API**: Used to link your mattermost account to your Google account
+3. Go to the credentials section, and create a new OAuth 2.0 credentials
+    - Under **Application type** specify Web Application
+    - Under **Authorized redirect URIs** add `https://(MM_SITE_URL)/plugins/com.mattermost.gcal/oauth2/complete` replacing `MM_SITE_URL` with your Mattermost instace site URL.
```
mickmister, line 28:
```suggestion
    - Under **Authorized redirect URIs** add `https://(MM_SITE_URL)/plugins/com.mattermost.gcal/oauth2/complete` replacing `MM_SITE_URL` with your Mattermost instance site URL.
```

server/plugin/plugin.go
```diff
@@ -88,6 +89,8 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
+	log.Printf("%v\n", p.env)
```
mickmister, line 92: Note to remove this before merge

server/plugin/plugin.go
```diff
@@ -323,11 +326,11 @@ func (p *Plugin) initEnv(e *Env, pluginURL string) error {
...
 			&model.Bot{
-				Username:    config.BotUserName,
-				DisplayName: config.BotDisplayName,
-				Description: config.BotDescription,
+				Username:    e.Provider.BotUsername,
+				DisplayName: e.Provider.BotDisplayName,
+				Description: fmt.Sprintf(config.BotDescription, e.Provider.DisplayName),
 			},
-			filepath.Join("assets", "profile.png"),
+			filepath.Join("assets", fmt.Sprintf("profile-%s.png", e.Provider.DisplayName)),
```
mickmister, line 333: I think we want `Name` here for the icon, i.e. `gcal` and `mscalendar`

```suggestion
			filepath.Join("assets", fmt.Sprintf("profile-%s.png", e.Provider.Name)),
```

mickmister (approved):
LGTM :tada: Also thanks for the detailed PR description

For a follow up task, we'll want to make sure CI builds the plugin for both providers

PR #270, opened 2023-07-25
[GCAL] Added encrypted key value storage

fmartingr:
#### Summary
- Adds an encrypted key value storage using AES256-GCM
- Storage is only enabled if the new `ProviderSettings.EncryptedStorage` is set to `true`
- [x] Replace logic for hardcoded `EncryptedStorage`.
  - Pending: #268 
- Adds `EncryptionKey` generated setting.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53793


server/store/store.go
```diff
@@ -55,8 +55,13 @@ type pluginStore struct {
 	Tracker            tracker.Tracker
 }
 
-func NewPluginStore(api plugin.API, logger bot.Logger, tracker tracker.Tracker) Store {
+func NewPluginStore(api plugin.API, logger bot.Logger, tracker tracker.Tracker, enableEncryption bool, encryptionKey []byte) Store {
 	basicKV := kvstore.NewPluginStore(api)
+
+	if enableEncryption {
+		basicKV = kvstore.NewEncryptedKeyStore(basicKV, encryptionKey)
```
mickmister, line 62: Encrypting everything in the kv store can make things hard to debug - locally or with the customer. I suggest we just encrypt the OAuth tokens. @esarafianou Do you have any thoughts on this?

fmartingr: Yeah I just added the implementation, but we can enable it on individual stores as required. I'm assuming we won't be using it locally very much 🙃

esarafianou: Yes, encrypting the oauth tokens is enough. It's not required to encrypt the calendar data.

plugin.json
```diff
@@ -98,6 +98,12 @@
                 "type": "text",
                 "help_text": "This should look something like \"googlebd09b1075898e210.html\"",
                 "default": ""
+            },
+            {
+                "key": "EncryptionKey",
+                "display_name": "Encryption key",
+                "type": "generated",
+                "help_text": "**Required**: The encryption key used to store data in the database"
```
mickmister, line 106: We'll want to think about backwards compatibility here for the mscalendar plugin. When we introduce this in a plugin, we typically release a new major version, which then links the admin to release notes before upgrading the plugin, so they can see what the breaking change is.

Unfortunately this will require all of their users to reconnect. Regenerating this value results in the same situation of data that cannot be unencrypted. One solution we've used is to DM all connected users to reconnect their accounts.

server/plugin/plugin.go
```diff
@@ -166,6 +168,9 @@ func (p *Plugin) OnConfigurationChange() (err error) {
 	p.updateEnv(func(e *Env) {
 		p.initEnv(e, pluginURL)
 
+		// TODO: replace with proper logic
+		e.EncryptedStore = true
```
mickmister, line 172: Probably good to address this in this PR or create a ticket for this

fmartingr: Pending #268

plugin.json
mickmister: > We could also allow a sysadmin command to load, encrypt and store the data again, to alleviate the pain a bit.

If we can make it turnkey for the admin, then I'm up for this. Note that some cloud customers are using mscalendar, so having the admin do anything database-level is not possible there.

We could also just do a migration on plugin start up. Here's an example of a migration using the kv store and `cluster.Mutex` https://github.com/mattermost/mattermost-plugin-gitlab/pull/361. Note that the migration would be a bit different for the mscalendar plugin, in that we would just be encrypting the data, and not sending user DMs etc. that the linked migration is doing. We would also want to make it so an encryption secret is automatically assigned if it is blank on plugin start up.

server/plugin/plugin.go
fmartingr: I have made it so the google calendar plugin uses the encrypted store while the mscalendar doesn't until we have a decission on what to do for mscalendar users.

providers/gcal.go
```diff
@@ -20,5 +20,7 @@ func GetGcalProviderConfig() *config.ProviderConfig {
 
 		BotUsername:    ProviderGCal,
 		BotDisplayName: ProviderGCalDisplayName,
+
+		EncryptedStore: true,
```
mickmister, line 24: Can we add `EncryptedStore: false` in `msgraph` provider for readability?

server/plugin/plugin.go
```diff
@@ -113,8 +113,9 @@ func (p *Plugin) OnActivate() error {
 				telemetry.NewLogger(p.API),
 			),
 		)
+		// TODO: replace with proper logic
```
mickmister, line 116: We can remove this line

mickmister (approved):

PR #272, opened 2023-07-25
[GCAL] Enable notifications and reminders when a superuser token is not supported

fmartingr:
#### Summary

Review the notifications and subscription logic to make it work with Google Calendar.

I've modified around things to allow the `Sync`/`SyncAll`/`ProcessAllDailySummary` to work without super user tokens, which I think is the same problem right now with the Google Calendar integration.

The changes work in a way that, if the `MakeSuperUserClient` returns a specific error ("not supported"), we use the individual credentials to make the API calls.

This has a big downside, and that is we can't batch the requests to Google, since each request uses a different credential.

Also, a fun thing I found during this is that even if you remove your application from Google, you still receive the webhook notifications 🥲 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53796


server/command/command.go
```diff
@@ -107,7 +107,7 @@ func (c *Command) Handle() (string, bool, error) {
 		handler = c.requireConnectedUser(c.findMeetings)
 	case "showcals":
 		handler = c.requireConnectedUser(c.showCalendars)
-	case "availability":
+	case "avail":
```
fmartingr, line 110: I was getting tired of writing it so many times. 😅

server/mscalendar/availability.go
```diff
@@ -63,9 +71,84 @@ func (m *mscalendar) SyncAll() (string, *StatusSyncJobSummary, error) {
...
+		}
+		if user.Settings.UpdateStatus || user.Settings.ReceiveReminders {
+			users = append(users, user)
+		}
+
+		start := time.Now().UTC()
+		end := time.Now().UTC().Add(calendarViewTimeWindowSize)
+
+		calendarUser := newUserFromStoredUser(user)
+		calendar, err := m.GetCalendarEvents(calendarUser, start, end)
```
mickmister, line 120: `calendar` is a bit vague, maybe `events`, `calendarEvents`, `calendarView`?

server/mscalendar/filters.go
```diff
@@ -31,6 +31,13 @@ func withRemoteUser(user *User) func(m *mscalendar) error {
 	}
 }
 
+func withActingUser(mattermostUserID string) func(m *mscalendar) error {
```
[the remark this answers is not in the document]
mickmister: Not cheating per se, but maybe error-prone. This is being called in a service account context, so we are changing the state of `m` temporarily while handling a given user's events. This makes it difficult to do concurrent operations across users.

server/mscalendar/notification.go
```diff
@@ -124,7 +119,9 @@ func (processor *notificationProcessor) work() {
 func (processor *notificationProcessor) processNotification(n *remote.Notification) error {
 	sub, err := processor.Store.LoadSubscription(n.SubscriptionID)
 	if err != nil {
-		return err
+		// TODO: Only happening when the subs were removed locally but not on remote, I kept getting
+		// notifications even after removing the app manually from my google account page.
+		return nil
```
mickmister, line 124: For any todos related to gcal, I think having an indicator like "TODO gcal" would help us know where the todo came from, i.e. from the gcal migration. For instance, this could be a breaking change for mscalendar provider, though looks like risk is not too bad

server/mscalendar/notification_format.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/config"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/mscalendar/views"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/fields"
+
+	"github.com/mattermost/mattermost-server/v6/model"
+)
+
+func (processor *notificationProcessor) newSlackAttachment(n *remote.Notification) *model.SlackAttachment {
```
mickmister, line 16: Just wondering, did this file's contents change at all after copying them here?

server/mscalendar/subscription.go
```diff
@@ -115,15 +118,16 @@ func (m *mscalendar) DeleteMyEventSubscription() error {
...
+		return err
 	}
 
-	err = m.DeleteOrphanedSubscription(subscriptionID)
+	err = m.Store.DeleteUserSubscription(m.actingUser.User, subscriptionID)
 	if err != nil {
-		return err
+		return errors.WithMessagef(err, "failed to delete subscription %s", subscriptionID)
 	}
+
```
mickmister, line 130: Logic makes sense to me. Remote remote first, then local right?

server/remote/msgraph/event.go
```diff
@@ -49,3 +50,7 @@ func (c *client) TentativelyAcceptEvent(remoteUserID, eventID string) error {
 	}
 	return nil
 }
+
+func (c *client) GetEventsBetweenDates(_ string, start, end time.Time) ([]*remote.Event, error) {
+	return nil, errors.New("not implemented")
```
mickmister, line 55: How can we make sure this doesn't get called, both now and in the future?

server/mscalendar/daily_summary.go
```diff
@@ -92,6 +93,65 @@ func (m *mscalendar) SetDailySummaryEnabled(user *User, enable bool) (*store.Dai
 	return dsum, nil
 }
 
+func (m *mscalendar) processAllDailySummaryByUser(now time.Time, userIndex store.UserIndex) error {
```
mickmister, line 96: `ByUser` makes me think "for one user". Maybe `processAllDailySummaryWithIndividualCredentials`

mickmister (commented):
Awesome work on this :tada:

Just a few comments for now. Just wondering, are the notifications features now working properly from this PR?

server/mscalendar/filters.go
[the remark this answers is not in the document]
fmartingr: This is the only way I imagined of leaving the architecture as it is, without moving other parts of the logic outside of the main "calendar" logic, but I do not know how this could affect elsewhere, if anything. Only running under GCal at the moment. Open to suggestions!

server/mscalendar/notification.go
fmartingr: I had this only for local testing to avoid seeing error messages that didn't belong to what I was currenctly developing. I'm going to revert it to what it was.

server/mscalendar/notification_format.go
fmartingr: Not sure what you mean. I only moved the parts of the logic that were about formatting, this file didn't exist previously. And I didn't modify any code present here, it's a direct copy & paste.

(Though I will modify some of it to make it similar to the mockups in another PR.)

server/mscalendar/subscription.go
fmartingr: Yes, try to remove the remote first, and if it fails, do not remove the local. I got my test environment in a state where I failed to remove some remote subscription but the local data was removed previously, so I could track anything back.

server/remote/msgraph/event.go
fmartingr: Implementing it is the safest way.

server/remote/msgraph/get_default_calendar_view.go
```diff
@@ -30,16 +30,7 @@ type calendarViewBatchResponse struct {
...
-
-	res := &calendarViewResponse{}
-	err := c.rbuilder.Users().ID(remoteUserID).CalendarView().Request().JSONRequest(
-		c.ctx, http.MethodGet, paramStr, nil, res)
-	if err != nil {
-		return nil, errors.Wrap(err, "msgraph GetDefaultCalendarView")
-	}
-
-	return res.Value, nil
+	return c.GetEventsBetweenDates(remoteUserID, start, end)
```
mickmister, line 33: It seems like these two methods are just copy of each other. Maybe we should just use `GetDefaultCalendarView`? Or keep `GetEventsBetweenDates` as it's more descriptive. Do we need two different implementations for gcal?

server/mscalendar/daily_summary.go
```diff
@@ -92,6 +93,65 @@ func (m *mscalendar) SetDailySummaryEnabled(user *User, enable bool) (*store.Dai
 	return dsum, nil
 }
 
+func (m *mscalendar) processAllDailySummaryWithIndividualCredentials(now time.Time, userIndex store.UserIndex) error {
```
mickmister, line 96: We may be able to de-duplicate some code between `ProcessAllDailySummary` and `processAllDailySummaryWithIndividualCredentials`. It would also be good to get this code under a unit test

Also it seems this block is missing, that exists in `ProcessAllDailySummary`:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/429698a2ca4e3680fdf12cf502f2c3ae02af089b/server/mscalendar/daily_summary.go#L231-L232

server/mscalendar/daily_summary.go
```diff
@@ -102,6 +162,11 @@ func (m *mscalendar) ProcessAllDailySummary(now time.Time) error {
 	}
 
 	err = m.Filter(withSuperuserClient)
+	// Allow processing the daily summary using individual credentials for remotes that doesn't allow
```
mickmister, line 165: Minor nit:

```suggestion
	// Allow processing the daily summary using individual credentials for remotes that don't allow
```

server/mscalendar/availability.go
```diff
@@ -63,9 +71,84 @@ func (m *mscalendar) SyncAll() (string, *StatusSyncJobSummary, error) {
...
+		return m.syncUsersIndividually(userIndex)
+	}
+	if err != nil {
+		return "", &StatusSyncJobSummary{}, errors.Wrap(err, "not able to filter the super user client")
+	}
+
 	return m.syncUsers(userIndex)
 }
 
+func (m *mscalendar) syncUsersIndividually(userIndex store.UserIndex) (string, *StatusSyncJobSummary, error) {
```
mickmister, line 87: What do you think about trying to de-duplicate some code between `syncUsers` and  `syncUsersIndividually`? Probably good to have a unit test for this as well

server/mscalendar/availability.go
```diff
@@ -63,9 +71,84 @@ func (m *mscalendar) SyncAll() (string, *StatusSyncJobSummary, error) {
 		return "", &StatusSyncJobSummary{}, errors.Wrap(err, "not able to load the users from user index")
 	}
 
+	err = m.Filter(withSuperuserClient)
```
mickmister, line 74: It's possible this could have side effects since this mutates state, and thus affects functions like `syncUsers`. From a quick look, it seems that it won't have issues though.

What's weird is that the only client used in that path is in `GetCalendarViews`, which is setting its own user-level client for batch requests. Not sure how that is working this way:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/d78381aa8bf00887bab751a9d2a33d7571c6c5e5/server/mscalendar/availability.go#L388-L389

server/mscalendar/notification_format.go
mickmister: > And I didn't modify any code present here, it's a direct copy & paste.

This was actually what I was wondering. LGTM :+1:

mickmister (changes requested):
Looking very good :+1:  Some comments for discussion, and request to de-duplicate some code, and add some unit tests for the new functionality

server/remote/msgraph/get_default_calendar_view.go
fmartingr: In my head, we should have a `GetEvents` that accepts filtering (right now it is this `GetEventsBetweenDates`), and then having a `views` file with custom methods for specific defined views (like the calendar, today, tomorrow ones) or either move that methods to the main "core" implementation. But I'm leaving that as a nice-to-have, since it doesn't specifically targets something related to the current task.

server/mscalendar/filters.go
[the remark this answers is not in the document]
mickmister: In general this seems fine, though just concerned about potential pitfalls of mutating the state. Really that's just an artifact of the `.Filter()` method in general. It's mainly meant to be used in a per-user context, but we introduced the superuser token later. And now we're doing sort of a hybrid, where we're using user-level OAuth tokens in a scheduled job context.

I think it's fine how this PR implements it, though we will need to rethink this if we want to perform these requests concurrently across different users. Speaking of which, what are your thoughts on the performance of the current approach, versus fanning out the requests in some fashion? The timing of reminders etc. depends on the speed of this operation, so I'm thinking we'll need some performance optimization here in the case of sending individual requests for each user.

server/remote/msgraph/get_default_calendar_view.go
mickmister: I suppose what I'm concerned about is that these are generally doing the same thing right now, though the implementation is slightly different (meaning not direct copy & paste, though seems to do the same high-level operation). Any reason not to have them share functionality by having this method call the existing one? Mainly trying to avoid duplicated code getting out of sync. I trust the direction you have here.

fmartingr: Assuming you mean "this method = `GetEventsBetweenDates`", since `GetDefaultCalendarView` is just a view with a predefined filter/paremeters, I think this is the one that should call the sightly lower-level method that actually retrieves the data, since the "view" only specifies parameters in my head, the same way a `GetTodayView`, `GetTomorrowView` would for other commands.

server/mscalendar/filters.go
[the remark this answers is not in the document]
fmartingr: In order to parallelize work we can do a `Filter` function that returns a copy of the `mscalendar` with the context switched so we can send out all the jobs to goroutines and wait for all of them to finish using a waitgroup. Is not the best approach, but it surely will help.

mickmister (approved):
LGTM :+1: just a few comments

As an aside, if possible I would avoid force-pushing since this makes it so GitHub's UI shows the PR commit history out of line with comments made before the most recent push.

mickmister:
Great work on this btw!! Awesome to see this implemented :tada:

server/mscalendar/availability_test.go
```diff
@@ -328,6 +352,86 @@ func TestReminders(t *testing.T) {
...
+		s.EXPECT().LoadUser(testUser.MattermostUserID).Return(testUser, nil)
+
+		m := New(e, "").(*mscalendar)
+		jobSummary := &StatusSyncJobSummary{}
+
+		_, _, err := m.retrieveUsersToSync(userIndex, jobSummary, true)
+		require.Error(t, err, errNoUsersNeedToBeSynced)
+	})
+
+	t.Run("one user should be synced", func(t *testing.T) {
```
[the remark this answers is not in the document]
fmartingr: Added two more cases, discovered a bug in the process :)

server/mscalendar/availability_test.go
```diff
@@ -430,6 +430,105 @@ func TestRetrieveUsersToSyncIndividually(t *testing.T) {
...
+		c.EXPECT().GetEventsBetweenDates(testUser.Remote.ID, gomock.Any(), gomock.Any()).Return(eventsUser1, nil)
+		c.EXPECT().GetEventsBetweenDates(testUser2.Remote.ID, gomock.Any(), gomock.Any()).Return(eventsUser2, nil)
+
+		m := New(e, "").(*mscalendar)
+		jobSummary := &StatusSyncJobSummary{}
+
+		users, responses, err := m.retrieveUsersToSync(userIndex, jobSummary, true)
+		require.NoError(t, err)
+		require.Equal(t, users, []*store.User{testUser, testUser2})
+		require.Equal(t, responses, []*remote.ViewCalendarResponse{{
```
mickmister, line 524: Nit: The order of args should go `t, expected, actual` rather than `t, actual, expected`. This makes it so a failing test output will display the correct error message

server/mscalendar/availability.go
```diff
@@ -441,12 +447,7 @@ func (m *mscalendar) setStatusOrAskUser(user *store.User, currentStatus *model.S
 }
 
 func (m *mscalendar) GetCalendarEvents(user *User, start, end time.Time) (*remote.ViewCalendarResponse, error) {
-	err := m.Filter(withActingUser(user.MattermostUserID))
-	if err != nil {
-		return nil, errors.Wrap(err, "error withRemoteUser")
-	}
-
-	err = m.Filter(withClient)
+	err := m.Filter(withClient)
```
mickmister, line 450: Do you see an opportunity to make this code less error-prone? It concerns me that it's easy to make an error based on where this call is placed

mickmister (approved):

server/mscalendar/availability_test.go
fmartingr: I always put the order wrong. 🤦

server/mscalendar/availability.go
fmartingr: That was the intention when I removed the `withActingUser` from this method. Now we can call this one without problems from other places since the new `withActingUser` only gets called from the fenced code in the "main" calls.

mickmister: Right, but what can we do to avoid needing to have `withActingUser` in the fenced code? Just want to make sure we don't run into something similar later, especially if it's something not caught by a unit test

PR #273, opened 2023-07-27
[GCAL] today/tomorrow commands with styling

fmartingr:
#### Summary
- Adds: `/trigger summary today`, `/trigger summary tomorrow`
- Aliases: `/trigger today`, `/trigger tomorrow`.
- Refactored `GetDailySummaryForUser(user *User) (string, error)` to `GetDaySummaryForUser(now time.Time, user *User) (string, error)` (allows sending the day, since the logic is the same).
- **Views:** Added new `RenderDaySummary` which uses slack attachments to display each event (see screenshots below)  
- **Poster:** Added new `DMWithMessageAndAttachments(mattermostUserID, message string, attachments ...*model.SlackAttachment) (string, error)`

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-53818
- https://mattermost.atlassian.net/browse/MM-53820

#### Notes

- I have not included the button because we can't directly add an URL in there, we need to add a handler. It also didn't make much sense since we have the link just above the button anyway. (Discussion pending)
- The views could be interfaced to allow the users to select which view they prefect. Personally I like the table view, but having the option to choose is nice. I didn't want to spend time on that just yet.
- The `Poster` cloud be improved by using a more generative approach: Example:`Poster.DM(mattermostUserID, option.WithAttachments(attachments), options.WithMessage(message))`.

#### Screenshots

Screenshot 2023-07-27 at 10 16 33


mickmister:
> Personally I like the table view, but having the option to choose is nice.

I have to agree 🙂  The attachments version is a little difficult to visually at a glance. I think a structured view like a table works better here. Are you able to post a screenshot of the table version for a before & after comparison?

0/5 on if it should be the table view, and/or have a toggle for this. @DHaussermann I'm curious what you think about this?

server/mscalendar/views/calendar.go
```diff
@@ -41,6 +43,67 @@ func RenderCalendarView(events []*remote.Event, timeZone string) (string, error)
...
+			Text:    fmt.Sprintf("(%s - %s)", event.Start.In(timezone).Time().Format(time.Kitchen), event.End.In(timezone).Time().Format(time.Kitchen)),
+			Fields:  fields,
+			Actions: actions,
+		})
+	}
+
+	return message, attachments, nil
+}
+
+// func getActionForLocation(loc *remote.Location) (action *model.PostAction) {
```
mickmister, line 92: I'm thining we can remove this commented code for now. We can visit this PR to grab this code if necessary. What do you think @fmartingr?

mickmister (approved):
Code LGTM :+1:  Just a few comments

fmartingr:
## Table view
Screenshot 2023-07-29 at 07 45 01

## Attachment view
Screenshot 2023-07-29 at 07 46 52

mickmister:
Looking at the before & after (table vs attachment view), I definitely prefer the table view. I think the attachment view is good for showing one event's information, but not so much when trying to visually parse several events at once.

mickmister (approved):
LGTM, though I would like to see the table view supported for the daily summary. @matthewbirtch Do you have any thoughts on this?

jasonblais:
> Looking at the before & after (table vs attachment view), I definitely prefer the table view. I think the attachment view is good for showing one event's information, but not so much when trying to visually parse several events at once.

+1 for a table view for a more concise daily summary of events

matthewbirtch:
> LGTM, though I would like to see the table view supported for the daily summary. @matthewbirtch Do you have any thoughts on this?

I like the table view for the agenda too. It's quite easy to scan and consume the agenda. A couple questions:
- Would we add a column for location? 
- What does the linked subject line go to?
- Minor detail: are we able to left-align the `Time` header cell? It should be left-aligned rather than centered. (like the Subject header cell).

mickmister:
> > * One possible concern for the table view would be mobile though. Displaying this on mobile wouldn't be great. Is it possible to display the other layout on mobile?

> I'm deferring to @mickmister on this one, since I don't think we can but I'm 0/5 on it.

@fmartingr Are you able to post a screenshot of what this looks like on mobile? I'm also pretty sure that we aren't able to display it differently on mobile

fmartingr:
> > > * One possible concern for the table view would be mobile though. Displaying this on mobile wouldn't be great. Is it possible to display the other layout on mobile?
> 
> > I'm deferring to @mickmister on this one, since I don't think we can but I'm 0/5 on it.
> 
> @fmartingr Are you able to post a screenshot of what this looks like on mobile? I'm also pretty sure that we aren't able to display it differently on mobile

## Message
IMG_5519

## When you tap on the lower-right icon
IMG_5520

cc: @larkox

fmartingr:
This is how it looks like on web with the "Join Meeting" cells:

Screenshot 2023-07-31 at 18 03 53

server/command/daily_summary.go
```diff
@@ -29,11 +29,11 @@ func (c *Command) dailySummary(parameters ...string) (string, bool, error) {
 		}
 		return postStr, false, nil
 	case "tomorrow":
-		_, err := c.MSCalendar.GetDaySummaryForUser(time.Now().Add(time.Hour*24), c.user())
+		postStr, err := c.MSCalendar.GetDaySummaryForUser(time.Now().Add((time.Hour*24)*4), c.user())
```
mickmister, line 32: Why do we add `(time.Hour*24)*4` here? Also do we need to take into account weekends (maybe that's the intention here)?

server/mscalendar/views/calendar.go
```diff
@@ -93,17 +94,26 @@ func renderEvent(event *remote.Event, asRow bool, timeZone string) (string, erro
...
+	}
+
 	subject := EnsureSubject(event.Subject)
 
-	return fmt.Sprintf(format, start, end, subject, link), nil
+	return fmt.Sprintf(format, start, end, subject, link, other), nil
+}
+
+func isKnownMeetingURL(location string) bool {
+	return strings.Contains(location, "zoom.us/j/") || strings.Contains(location, "discord.gg") || strings.Contains(location, "meet.google.com")
```
mickmister, line 116: Neat :+1:  Something comes to mind, that a customer may have other call-related domains, including Mattermost Calls. Maybe have this configurable in some way?

mickmister (commented):
Just one question regarding `tomorrow` date calculation. Otherwise LGTM :+1:

server/command/daily_summary.go
fmartingr: Oh no no no, I accidentaly commited that. Used to get the output of a specific day for a screenshot.

server/mscalendar/views/calendar.go
fmartingr: Yeah, I want to integrate calls too, but need to send the SITE_URL over to check the URL to do it properly, so it needs a bit more work. Also, Google sends `conferenceData` for some events instead of location (I'm assuming they have a specific integration), so it wont work every time.

server/command/daily_summary.go
mickmister: Though this begs the question, should we have a special case for weekends? Currently, the daily summary job avoids sending the summary on the weekend, so there is already some convention in the plugin to avoid weekends. Maybe the `tomorrow` command should do something similar?

fmartingr: Didn't check that code, is it configurable? Does it take into consideration countries were weekends are longer/shorter/on different days?

mickmister: Here is the line that does this https://github.com/mattermost/mattermost-plugin-mscalendar/blob/f226b5e7b2620c1ab625a2f2e4a31cbe6c1a6b30/server/mscalendar/daily_summary.go#L270

It's currently not configurable, but we can use `WorkingHours` to figure out the correct days if available for gcal:

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/65b2e6734e76a6384a71101d809377f8b104aa62/server/remote/user.go#L13-L25

PR #274, opened 2023-07-27
[GCAL] Channels reminder underlying logic

fmartingr:
#### Summary

Adds new functionality to `MSCalendar.notifyUpcomingEvents` method by checking for event metadata on the KVStore, if the event is present and have `LinkedChannels` set, we will post a reminder in all channels present too, the same way we do in DMs.
- [x] Check recurrent events
- [x] Create methods to CRUD this information in the store.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53843


server/store/event_store.go
```diff
@@ -18,18 +20,29 @@ import (
 const ttlAfterEventEnd = 30 * 24 * time.Hour // 30 days
 const defaultEventTTL = 30 * 24 * time.Hour  // 30 days
 
+type EventMetadata struct {
+	LinkedChannels map[string]struct{}
```
mickmister, line 24: Maybe put a doc comment on this field? Conceptually I expected this to be a `[]string` to store the channel ids, but I understand it's a map so it's easier to use for lookup and/or forcing uniqueness. Maybe name it `LinkedChannelIds` to show that there is no other data involved?

Originally when reading the line below, I thought there was a bug where we were accidentally using the index of an array and mistaking that for a channel id, as I thought this was an array before checking the `EventMetadata` struct definition. https://github.com/mattermost/mattermost-plugin-mscalendar/blob/fced1661a8b20065c01e97acacf6944e0ee16486/server/mscalendar/availability.go#L531

server/mscalendar/availability_test.go
```diff
@@ -293,6 +295,30 @@ func TestReminders(t *testing.T) {
 			numReminders:   2,
 			shouldLogError: false,
 		},
+		"Remote event linked to channel in the range for the reminder. DM and channel reminders should occur.": {
+			remoteEvents: []*remote.Event{
+				{ID: "event_id", ICalUID: "event_id", Start: remote.NewDateTime(time.Now().Add(7*time.Minute).UTC(), "UTC"), End: remote.NewDateTime(time.Now().Add(45*time.Minute).UTC(), "UTC")},
+			},
+			eventMetadata: map[string]*store.EventMetadata{
+				"event_id": {
+					LinkedChannels: map[string]struct{}{"channel_id": {}},
```
mickmister, line 304:
```suggestion
					LinkedChannels: map[string]struct{}{"some_channel_id": {}},
```

Just to differentiate from `channel_id` potentially being used as a namespace

server/mscalendar/availability_test.go
```diff
@@ -293,6 +295,30 @@ func TestReminders(t *testing.T) {
...
+				},
+			},
+			numReminders:   1,
+			shouldLogError: false,
+		},
+		"Remote recurring event linked to channel in the range for the reminder. DM and channel reminders should occur.": {
+			remoteEvents: []*remote.Event{
+				{ID: "event_id_20230728", RecurringEventID: "event_id", ICalUID: "event_id", Start: remote.NewDateTime(time.Now().Add(7*time.Minute).UTC(), "UTC"), End: remote.NewDateTime(time.Now().Add(45*time.Minute).UTC(), "UTC")},
```
mickmister, line 312: Can we set the three event id fields to different values, if they are typically different values? Just to help catch issues if we change logic around using the fields. Same for the test case above

server/remote/event.go
```diff
@@ -26,6 +27,13 @@ type Event struct {
 	ResponseRequested          bool                 `json:"responseRequested,omitempty"`
 }
 
+func (e Event) GetMainID() string {
+	if e.RecurringEventID != "" {
+		return e.RecurringEventID
+	}
+	return e.ID
```
mickmister, line 34: Any reason to use `ID` over `ICalUID`? IIRC `ICalUID` is a better unique identifier since it's the same across all users' calendars, though there are different use cases for using one over the other

fmartingr: I was pretty confused on why there were so many IDs for an event. But yes, using ICalUID should work in this scenario.

PR #275, opened 2023-07-28
[GCAL] Endpoint to autocomplete connected calendar users

fmartingr:
#### Summary

- Adds a `/autocomplete/users` endpoint to the plugin to search for connected users by username, display name and email.
  - [x] Maybe don't expose the entire `UserShort` struct in the results?
- Adds new `UserStore` method:`SearchInUserIndex(term string, limit int) ([]UserShort, error)`
- Expands the `UserShort` struct with the username and display name from Mattermost.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53789


server/api/autocomplete.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/store"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/httputils"
+)
+
+func (api *api) autocompleteUsers(w http.ResponseWriter, r *http.Request) {
+	// TODO: This doesn't seem like proper authorization. Anyone can send a header to an http
+	// request impersonating the user (?)
+	mattermostUserID := r.Header.Get("X-Mattermost-User")
```
mickmister, line 17: I think this should be `Mattermost-User-Id` instead. Where does `X-Mattermost-User` come from?

```suggestion
	mattermostUserID := r.Header.Get("Mattermost-User-Id")
```

Check out these links for more info on why this is a trusted approach. If that header is defined, this is an authenticated request for that user. If it's not defined, it's not an authenticated request.

- https://developers.mattermost.com/integrate/plugins/components/server/reference/#Hooks.ServeHTTP
- https://github.com/mattermost/mattermost/blob/93a2c3281ad896a84fc2a3fd6a52708a765a7dd9/server/channels/app/plugin_requests.go#L153
- https://github.com/mattermost/mattermost/blob/93a2c3281ad896a84fc2a3fd6a52708a765a7dd9/server/channels/app/plugin_requests.go#L209

server/store/user_store.go
```diff
@@ -230,6 +239,26 @@ func (s *pluginStore) DeleteUserFromIndex(mattermostUserID string) error {
...
 
+func (s *pluginStore) SearchInUserIndex(term string, limit int) ([]UserShort, error) {
+	userIndex, err := s.LoadUserIndex()
+	if err != nil {
+		return nil, errors.Wrap(err, "error searching user in index")
+	}
+
+	result := make([]UserShort, 0)
+	for idx, u := range userIndex {
+		if strings.Contains(u.MattermostUsername, term) || strings.Contains(u.MattermostDisplayName, term) || strings.Contains(u.Email, term) {
```
mickmister, line 250: I wonder if the user would want to be able to invite arbitrary emails too. I suppose we can still let the user can do that, even the input doesn't match a connected user. The UI could just allow them to put whatever email they want.

@matthewbirtch Do you have any thoughts on this? We're implementing the "Attendees" selector in the create event modal. At the moment, the autocomplete searches for connected users, and find matches by comparing the autocomplete input with a user's username, display name in Mattermost, and connected email address. If any of them match, the user shows in the selector.

I would normally suggest we just use the Mattermost user autocomplete for this kind of thing, but I think we want to limit the results to connected users so we know their calendar email address. We could even try and use gcal's API for this autocomplete, but that is more complex and requires rework for compatibility with mscalendar.

The Figma design currently shows the Mattermost display name in the selected options. Would the autocomplete options be shown the same way? 0/5 I'm thinking the email address is significant here, since that's what is used on the calendar side.

CleanShot 2023-07-28 at 19 13 19@2x

mickmister (commented):
LGTM, just some questions for discussion

server/api/autocomplete.go
fmartingr: Honestly I got it from here: https://developers.mattermost.com/integrate/plugins/components/server/best-practices/#how-can-plugins-make-sure-http-requests-are-authentic but I don't know why I implemented it with another name 🤔

server/store/user_store.go
fmartingr: Yeah that's why I made the code find by everything, but we only have emails (thus, can use as attendees) of connected users. But I think that we can let it use arbitrary emails too. I will make some tests of that next week.

server/api/autocomplete.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/store"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/bot"
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/httputils"
+)
+
+func (api *api) autocompleteUsers(w http.ResponseWriter, r *http.Request) {
+	// TODO: This doesn't seem like proper authorization. Anyone can send a header to an http
+	// request impersonating the user (?)
```
mickmister, line 16:
```suggestion
```

mickmister (approved):
LGTM 🚢

server/store/user_store.go
matthewbirtch: I agree, that would be great to be able to invite arbitrary email addresses in addition to connected users. We do this in the 'Invite people to team' modal right now by adding an option in the autocomplete to invite via email

Autocomplete displays like this:
image

Chips display like this
image

Is it possible to leverage this behavior?

server/store/user_store.go
```diff
@@ -239,16 +253,16 @@ func (s *pluginStore) DeleteUserFromIndex(mattermostUserID string) error {
...
 
-func (s *pluginStore) SearchInUserIndex(term string, limit int) ([]UserShort, error) {
+func (s *pluginStore) SearchInUserIndex(term string, limit int) (UserIndex, error) {
 	userIndex, err := s.LoadUserIndex()
 	if err != nil {
 		return nil, errors.Wrap(err, "error searching user in index")
 	}
 
-	result := make([]UserShort, 0)
+	result := make([]*UserShort, 0)
```
mickmister, line 262: I'm used to the syntax of the following, and I would prefer to stay consistent across the codebase (and other plugin codebases). Mainly a coding style reference though

```suggestion
	result := []*UserShort{}
```

server/store/user_store.go
```diff
@@ -268,6 +282,14 @@ func (s *pluginStore) StoreUserActiveEvents(mattermostUserID string, events []st
 	return kvstore.StoreJSON(s.userKV, mattermostUserID, u)
 }
 
+func (index UserIndex) ToDTO() (result []UserShortDTO) {
+	for _, u := range index {
+		result = append(result, u.ToDTO())
+	}
+
+	return
```
mickmister, line 290: Unnecessary return

fmartingr: Why?

mickmister: Oh I see the return value var is in the signature. I suppose I'm not used to seeing that too often

mickmister (approved):

PR #276, opened 2023-07-31
[GCAL] Refactor Microsoft Calendar references to provider references

fmartingr:
#### Summary

Refactor to use the `config.ProviderConfig` data to display text in the plugin, mainly the display name and the command trigger in help text around the commands and events.

### DIFF
--- a/server/utils/oauth2connect/oauth2_complete.go
+++ b/server/utils/oauth2connect/oauth2_complete.go
@@ -4,12 +4,13 @@
 package oauth2connect
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/httputils"
 )
 
-func (oa *oa) oauth2Complete(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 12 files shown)


server/utils/oauth2connect/oauth2_complete.go
```diff
@@ -4,12 +4,13 @@
...
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/mattermost/mattermost-plugin-mscalendar/server/utils/httputils"
 )
 
-func (oa *oa) oauth2Complete(w http.ResponseWriter, r *http.Request) {
+func (oa *oa) oauth2Complete(w http.ResponseWriter, r *http.Request, providerDisplayName string) {
```
mickmister, line 13: `oa` already has access to `config.ProviderConfig`, so I think we don't need the `providerDisplayName` parameter right?

mickmister (approved):
LGTM :+1:

server/utils/oauth2connect/oauth2_complete.go
fmartingr: Oh right, that was from a previous change.

PR #278, opened 2023-08-01
[GCAL] Remove unused code

fmartingr:
#### Summary
Removed code related to batch requests that is not in use by the Google Calendar plugin.


server/remote/gcal/get_schedule.go
```diff
@@ -0,0 +1,9 @@
+package gcal
+
+import (
+	"github.com/mattermost/mattermost-plugin-mscalendar/server/remote"
+)
+
+func (c *client) GetSchedule(_ []*remote.ScheduleUserInfo, _, _ *remote.DateTime, _ int) ([]*remote.ScheduleInformation, error) {
```
mickmister, line 7: This is actually dead code for mscalendar as well. It's implemented but not called anywhere IIRC

mickmister (approved):

server/remote/gcal/get_schedule.go
fmartingr: Yeah, I noticed that but wasn't sure if it was a feature in development or something, so I wanted to be safe about it.

PR #279, opened 2023-08-01
[GCAL] Move configuration readyness logic to remotes

fmartingr:
#### Summary

Refactored the logic that checks for the correct configuration to the remotes instead of doing it in the main plugin load, since each remote has different requirements now.

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -71,10 +71,8 @@ func (p *Plugin) OnActivate() error {
 		return errors.WithMessage(err, "failed to load plugin configuration")
 	}
 
-	if stored.OAuth2Authority == "" ||
-		stored.OAuth2ClientID == "" ||
-		stored.OAuth2ClientSecret == "" {
-		return errors.New("failed to configure: OAuth2 credentials to be set in the config")
+	if errConfig := p.env.Remote.CheckConfiguration(stored); errConfig != nil {
+		return errors.Wrap(errConfig, "failed to configure: ")
 	}
 
 	p.initEnv(&p.env, "")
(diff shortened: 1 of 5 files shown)


server/plugin/plugin.go
```diff
@@ -71,10 +71,8 @@ func (p *Plugin) OnActivate() error {
 		return errors.WithMessage(err, "failed to load plugin configuration")
 	}
 
-	if stored.OAuth2Authority == "" ||
-		stored.OAuth2ClientID == "" ||
-		stored.OAuth2ClientSecret == "" {
-		return errors.New("failed to configure: OAuth2 credentials to be set in the config")
+	if errConfig := p.env.Remote.CheckConfiguration(stored); errConfig != nil {
+		return errors.Wrap(errConfig, "failed to configure: ")
```
mickmister, line 75: `errors.Wrap` should already add the colon here

```suggestion
		return errors.Wrap(errConfig, "failed to configure")
```

server/remote/msgraph/remote.go
```diff
@@ -85,3 +86,11 @@ func (r *impl) NewOAuth2Config() *oauth2.Config {
 		Endpoint: microsoft.AzureADEndpoint(r.conf.OAuth2Authority),
 	}
 }
+
+func (r *impl) CheckConfiguration(cfg config.StoredConfig) error {
+	if cfg.OAuth2ClientID == "" || cfg.OAuth2ClientSecret == "" || cfg.OAuth2Authority == "" {
+		return fmt.Errorf("OAuth2 credentials to be set in the config")
```
mickmister, line 92: Nice to have: Probably good to name the config settings as they are shown in the plugin config page, e.g. `Azure Directory (tenant) ID`

mickmister (approved):

server/remote/msgraph/remote.go
fmartingr: I'd rather do this later on and focus on gcal related things at the moment. We can do that once we change the mscalendar` name to something different and address all in the same PR.

PR #282, opened 2023-08-02
[GCAL] Fix event notifications not working due to missing scope permissions plus internal state

fmartingr:
#### Summary
- In order to retrieve the timezone we need `CalendarScope`. We will need it later to retrieve calendars anyway.
- For `deliverReminders` to work properly, since we do not have a superuser credential around (and we now use the `FilterCopy` logic) we had to change it's internal logic as well.

### DIFF
--- a/server/mscalendar/availability.go
+++ b/server/mscalendar/availability.go
@@ -167,7 +167,7 @@ func (m *mscalendar) syncUsers(userIndex store.UserIndex, fetchIndividually bool
 		return err.Error(), syncJobSummary, errors.Wrapf(err, "error retrieving users to sync (individually=%v)", fetchIndividually)
 	}
 
-	m.deliverReminders(users, calendarViews)
+	m.deliverReminders(users, calendarViews, fetchIndividually)
 	out, numberOfUsersStatusChanged, numberOfUsersFailedStatusChanged, err := m.setUserStatuses(users, calendarViews)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/mscalendar/availability.go
```diff
@@ -212,7 +212,15 @@ func (m *mscalendar) deliverReminders(users []*store.User, calendarViews []*remo
 		}
 
 		mattermostUserID := usersByRemoteID[view.RemoteUserID].MattermostUserID
-		m.notifyUpcomingEvents(mattermostUserID, view.Events)
+		if fetchIndividually {
+			engine, err := m.FilterCopy(withActingUser(user.MattermostUserID))
+			if err != nil {
+				m.Logger.With(bot.LogContext{"err": err}).Errorf("error getting engine for user")
+			}
+			engine.notifyUpcomingEvents(mattermostUserID, view.Events)
```
mickmister, line 220: Should we return when this error happens?

mickmister (approved):
Nice catch! I recall you saying that you thought this had to do with scopes

fmartingr:
> Nice catch! I recall you saying that you thought this had to do with scopes

Yeah, without the `CalendarScope` we weren't able to retrieve the Timezone. Maybe I had that scope when I developed the notifications? I changed it several time under several contexts, so maybe I missed that.

server/mscalendar/availability.go
fmartingr: Yeah, I put a `continue` in there to avoid trying to notify that user. Good catch.

PR #287, opened 2023-08-02
[GCAL] Better logging information

fmartingr:
#### Summary

Add some more information to some log lines to address what's happening in the test server

### DIFF
--- a/server/mscalendar/daily_summary.go
+++ b/server/mscalendar/daily_summary.go
@@ -127,7 +127,7 @@ func (m *mscalendar) ProcessAllDailySummary(now time.Time) error {
 
 		shouldPost, shouldPostErr := shouldPostDailySummary(dsum, now)
 		if shouldPostErr != nil {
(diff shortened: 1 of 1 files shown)


server/mscalendar/daily_summary.go
```diff
@@ -127,7 +127,7 @@ func (m *mscalendar) ProcessAllDailySummary(now time.Time) error {
 
 		shouldPost, shouldPostErr := shouldPostDailySummary(dsum, now)
 		if shouldPostErr != nil {
-			m.Logger.Warnf("Error posting daily summary for user %s. err=%v", storeUser.MattermostUserID, shouldPostErr)
+			m.Logger.With(bot.LogContext{"mm_user_id": storeUser.MattermostUserID, "now": now.String()}).Warnf("Error checking daily summary should be posted")
```
mickmister, line 130: We're no longer logging `shouldPostError`. Is this intentional?

fmartingr: I missed that.

PR #289, opened 2023-08-03
[GCAL] Disable `GetNotificationData` since it's not currently working

fmartingr:
#### Summary
Disable `GetNotificationData` since it's not currently working


fmartingr:
> @fmartingr I'm not sure what value this brings at the moment. Does the error just make it explicit that it's not working for the time being? I'm fine with merging though

It removes logic that does not work and show relevant information in the logs. THough this won't be running for Gcal with #292

PR #292, opened 2023-08-04
[GCAL] Event notifications behind a feature flag

fmartingr:
#### Summary

Only enable event notifications if the provider allows it, along with any relevant settings and wizard option

The subscribe/unsubscribe commands are removed from autocompletion and only admin can use them from now on.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53918

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -200,10 +200,12 @@ func (p *Plugin) OnConfigurationChange() (err error) {
 		welcomeFlow := mscalendar.NewWelcomeFlow(e.bot, e.Dependencies.Welcomer, e.Provider.Features)
 		e.bot.RegisterFlow(welcomeFlow, mscalendarBot)
 
-		if e.notificationProcessor == nil && e.Provider.Features.EventNotifications {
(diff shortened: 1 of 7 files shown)


server/plugin/plugin.go
```diff
@@ -194,12 +194,13 @@ func (p *Plugin) OnConfigurationChange() (err error) {
...
 
-		welcomeFlow := mscalendar.NewWelcomeFlow(e.bot, e.Dependencies.Welcomer)
+		welcomeFlow := mscalendar.NewWelcomeFlow(e.bot, e.Dependencies.Welcomer, e.Provider.Features)
 		e.bot.RegisterFlow(welcomeFlow, mscalendarBot)
 
-		if e.notificationProcessor == nil {
+		if e.notificationProcessor == nil && e.Provider.Features.EventNotifications {
 			e.notificationProcessor = mscalendar.NewNotificationProcessor(e.Env)
 		} else {
 			e.notificationProcessor.Configure(e.Env)
```
mickmister, line 206: Should both of these be wrapped in `if e.Provider.Features.EventNotifications`?

PR #295, opened 2023-08-04
[GCAL/MSCAL] Remind only accepted events

fmartingr:
#### Summary
Prevent event reminders from events that are not accepted.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53900
https://mattermost.atlassian.net/browse/MM-53983


server/remote/gcal/get_default_calendar_view.go
```diff
@@ -21,22 +21,17 @@ const (
...
 
 var responseStatusConversion = map[string]string{
-	GoogleResponseStatusYes:   ResponseYes,
-	GoogleResponseStatusMaybe: ResponseMaybe,
-	GoogleResponseStatusNo:    ResponseNo,
-	GoogleResponseStatusNone:  ResponseNone,
+	GoogleResponseStatusYes:   remote.EventResponseStatusAccepted,
+	GoogleResponseStatusMaybe: remote.EventResponseStatusTentative,
+	GoogleResponseStatusNo:    remote.EventResponseStatusDeclined,
+	GoogleResponseStatusNone:  remote.EventResponseStatusNotAnswered,
```
mickmister, line 34: Does this break compatibility with mscalendar in anyway?

mickmister (approved):

server/remote/gcal/get_default_calendar_view.go
fmartingr: It shouldn't since I didn't modify the `mscalendar` package nor the `remote.Event` struct.

mickmister: Before this change, both provider cases shared the same enum of values in `remote.Event.ResponseStatus`, but now the one returned by the `remote/gcal` package is different than the `remote/mscalendar` package. To be consistent, shouldn't we also change the `mscalendar` case, to use this new set of constants?

mickmister: It's not being _used_ in `msgraph`, but it is implicitly being assigned through the JSON decoding of the msgraph API response. My only concern is that there is core business logic in the plugin that may not match up with what provided by the `remote/msgraph` package, through the msgraph API.

fmartingr: We can make it better before we fully merge both plugins, I did a workaround under https://github.com/mattermost/mattermost-plugin-mscalendar/pull/295/commits/4a8d44be35b3d2ac182fae39ea5c66433a61522b

PR #296, opened 2023-08-10
[GCAL/MSCAL] Store linked events per user to remove links on disconnection

fmartingr:
#### Summary

This PR adds logic so each user has stored the events linked to channels as an index so when the user is disconnected all links are removed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53981


server/api/events.go
```diff
@@ -230,6 +230,12 @@ func (api *api) createEvent(w http.ResponseWriter, r *http.Request) {
 
 	// Event linking
 	if payload.ChannelID != "" {
+		if err := api.Store.StoreUserLinkedEvent(user.MattermostUserID, event.ICalUID, payload.ChannelID); err != nil {
+			api.Poster.DM(mattermostUserID, "You event **%s** could not be linked to a channel. Please contact an administrator for more details.", event.Subject)
```
mickmister, line 234:
```suggestion
			api.Poster.DM(mattermostUserID, "Your event **%s** could not be linked to a channel. Please contact an administrator for more details.", event.Subject)
```

server/mscalendar/user.go
```diff
@@ -137,6 +138,25 @@ func (m *mscalendar) DisconnectUser(mattermostUserID string) error {
...
+	if len(linkedEventsLeft) != 0 {
+		storedUser.ChannelEvents = linkedEventsLeft
+		if errStore := m.Store.StoreUser(storedUser); errStore != nil {
+			m.Logger.With(bot.LogContext{
+				"err":                  errStore,
+				"mm_user_id":           storedUser.MattermostDisplayName,
+				"linked_channels_left": linkedEventsLeft,
+			}).Errorf("error storing user after failing deleting linked channels from store")
+		}
+		return fmt.Errorf("error deleting linked channels from events")
```
mickmister, line 157: This should be in the `if errStore != nil` block right?

fmartingr: Not really, this will show an error either way if there are [events pending from removal](https://github.com/mattermost/mattermost-plugin-mscalendar/pull/296/files#diff-2c556af43018651b32cd1e69d78e9654556352a9ccc7446aa398a7884486f1f8R143-R147) in the previous block. Note how we only log the error on store but actually return an error on that condition.

- We try to remove events, if an event fails deletion (unlink) we add it to a `map`.
- After that, we check the `map`, if it contains elements, we failed to remove something we should **return** `error`.
  - While in this condition, we store the linked channels that are pending removal back to the user, and if that fails the **log** an error.

PR #297, opened 2023-08-14
[GCAL] Encrypt user store to store OAuth tokens safely

fmartingr:
#### Summary

Encrypt the user store since it contains the user's oauth tokens

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53852


mickmister (approved):

fmartingr:
> Maybe we should make it so encryption is disabled when `ServiceSettings.EnableDeveloper` is true?

While is a good feature to have locally, we would need to explain this very well in documentation for customers to not mess up plugin storage by toggling the developer setting in Mattermost. If they do for any reason (debugging, etc) they will break the plugin.

PR #303, opened 2023-08-18
[GCAL/MSCAL] Create event modal only for connected accounts

fmartingr:
#### Summary

Prevents non-connected users from opening the create event modal, both from the slash command (soon to be removed) and the channel header menu.

@mickmister how can I properly retrieve data from the backend and store it in the frontend for its use around? I exposed the provider data (to customize the `sendEphemeralMessage` call) but I'm not sure if creating a variable and populating it with the data I get from the backend is appropriate.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54114

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -30,7 +30,12 @@ export default class Plugin {
 
         registry.registerChannelHeaderMenuAction(
             <span><i className='icon fa fa-calendar-plus-o'/>{'Create calendar event'}</span>,
-            (channelID) => store.dispatch(openCreateEventModal(channelID)),
+            async (channelID) => {
+                if (await hooks.CheckUserIsConnected()) {
+                    store.dispatch(openCreateEventModal(channelID));
+                    return
+                }
+            }
         );
 
         // reminder to set up site url for any API calls
--- a/webapp/src/plugin_hooks.ts
+++ b/webapp/src/plugin_hooks.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {openCreateEventModal} from './actions';
+import {getConnected, openCreateEventModal, sendEphemeralPost} from './actions';
 import {isUserConnected} from './selectors';
 
 // import {openCreateModalWithoutPost, openChannelSettings, sendEphemeralPost, openDisconnectModal, handleConnectFlow, getConnected} from '../actions';
@@ -52,12 +52,10 @@ export default class Hooks {
     };
 
     checkUserIsConnected = async (): Promise<boolean> => {
-        return true;
-
         if (!isUserConnected(this.store.getState())) {
             await this.store.dispatch(getConnected());
             if (!isUserConnected(this.store.getState())) {
-                this.store.dispatch(sendEphemeralPost('Your Mattermost account is not connected to Jira. Please use `/jira connect` to connect your account, then try again.'));
+                this.store.dispatch(sendEphemeralPost('Your Mattermost account is not connected.'));
                 return false;
             }
         }
--- a/webapp/src/plugin_hooks.ts
+++ b/webapp/src/plugin_hooks.ts
@@ -43,7 +43,7 @@ export default class Hooks {
     };
 
     handleCreateEventSlashCommand = async (message: string, contextArgs: ContextArgs) => {
-        if (!(await this.checkUserIsConnected())) {
+        if (!(await this.CheckUserIsConnected())) {
             return Promise.resolve({});
         }
 
@@ -51,7 +51,7 @@ export default class Hooks {
         return Promise.resolve({});
     };
 
-    checkUserIsConnected = async (): Promise<boolean> => {
+    CheckUserIsConnected = async (): Promise<boolean> => {
         if (!isUserConnected(this.store.getState())) {
             await this.store.dispatch(getConnected());
             if (!isUserConnected(this.store.getState())) {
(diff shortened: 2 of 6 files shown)


mickmister:
> slash command (soon to be removed)

Thoughts on keeping the slash command (and rename to something like `/gcal event create`)? I think this would be a good way to schedule a meeting in the context of a thread

fmartingr:
> > slash command (soon to be removed)
> 
> Thoughts on keeping the slash command (and rename to something like `/gcal event create`)? I think this would be a good way to schedule a meeting in the context of a thread

I don't mind having the slash command, what seems weird to me is the frontend/backend handling commands independently from each other.

mickmister:
> @mickmister how can I properly retrieve data from the backend and store it in the frontend for its use around?

During the initialization of the plugin, we can first ask the server for the user's plugin connection status. Here is an example from the Jira plugin: https://github.com/mattermost/mattermost-plugin-jira/blob/17665067caaf92bd8ffff6719edb6ee36dfdb534/webapp/src/plugin.tsx#L31

That action also stores the result in the frontend's redux store for React components to show (also happening in this plugin currently)

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/127b79c52530f42c9229d9789d98e01810767e61/webapp/src/actions.ts#L126-L129

Then you can access from the selector here

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/127b79c52530f42c9229d9789d98e01810767e61/webapp/src/selectors.ts#L46

---

Note that we'll also want to use a websocket message to communicate to the frontend when the user gets connected/disconnected:

https://github.com/mattermost/mattermost-plugin-jira/blob/17665067caaf92bd8ffff6719edb6ee36dfdb534/webapp/src/plugin.tsx#L47

https://github.com/mattermost/mattermost-plugin-jira/blob/17665067caaf92bd8ffff6719edb6ee36dfdb534/server/user.go#L229-L231

---

In order to show/hide the channel menu component based on the connected status, we can do one of two things:

- Simply return `null` in that component when the user is not connected. This can result in UI implications, but I would try this first as it's very simple
- Use [unregisterComponent](https://developers.mattermost.com/integrate/plugins/components/webapp/reference/#unregisterComponent) to dynamically show/hide the channel header component

---

Also note that the initial connection check should run inside of the `setup` function defined in `initialize`. The point of the `setup` function is "run this once the user is logged in". Otherwise, the `initialize` code just runs on page load when the user is at the MM login screen.

https://github.com/mattermost/mattermost-plugin-mscalendar/blob/127b79c52530f42c9229d9789d98e01810767e61/webapp/src/index.tsx#L25

webapp/src/index.tsx
```diff
@@ -30,7 +30,11 @@ export default class Plugin {
 
         registry.registerChannelHeaderMenuAction(
             <span><i className='icon fa fa-calendar-plus-o'/>{'Create calendar event'}</span>,
-            (channelID) => store.dispatch(openCreateEventModal(channelID)),
+            async (channelID) => {
+                if (await hooks.checkUserIsConnected()) {
+                    store.dispatch(openCreateEventModal(channelID));
+                }
```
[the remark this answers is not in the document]
fmartingr: Where should I put it?

webapp/src/index.tsx
```diff
@@ -22,8 +22,14 @@ export default class Plugin {
         const hooks = new Hooks(store);
         registry.registerSlashCommandWillBePostedHook(hooks.slashCommandWillBePostedHook);
 
-        const setup = () => {
+        const setup = async () => {
+            // Retrieve provider configuration so we can access names and other options in messages to use in the frontend.
+            await store.dispatch(getProviderConfiguration());
```
mickmister, line 27: We also want to move the `registerChannelHeaderMenuAction` into this `setup` function, unless we're returning null

webapp/src/plugin_hooks.ts
```diff
@@ -55,7 +50,8 @@ export default class Hooks {
         if (!isUserConnected(this.store.getState())) {
             await this.store.dispatch(getConnected());
             if (!isUserConnected(this.store.getState())) {
-                this.store.dispatch(sendEphemeralPost('Your Mattermost account is not connected.'));
+                const providerConfiguration = await getProviderConfiguration(this.store.getState());
+                this.store.dispatch(sendEphemeralPost(`Your Mattermost account is not connected to ${providerConfiguration.DisplayName}. In order to create a calendar event please connect your account first using \`/${providerConfiguration.CommandTrigger} connect\`.`));
```
mickmister, line 54: We typically use snake case for the JSON serialization of things, but 0/5 here

mickmister (commented):
Generally LGTM, I would like to see the `Create calendar event` menu action only visible upon being connected. Alternatively, we could show an ephemeral message when it's clicked, if the user is not connected.

fmartingr:
> Alternatively, we could show an ephemeral message when it's clicked, if the user is not connected.

This is what I ended up doing because I think it helps with discoverability.

webapp/src/plugin_hooks.ts
fmartingr: I didn't bother since this wasn't intended to be public, but I can change it.

webapp/src/plugin_hooks.ts
```diff
@@ -30,7 +30,8 @@ export default class Hooks {
             return Promise.resolve({message, args: contextArgs});
         }
 
-        if (message.startsWith(createEventCommand)) {
+        const providerConfiguration = await getProviderConfiguration(this.store.getState());
```
mickmister, line 33: `getProviderConfiguration` is a synchronous function so no need to await here

```suggestion
        const providerConfiguration = getProviderConfiguration(this.store.getState());
```

mickmister (approved):

PR #304, opened 2023-08-18
#WIP [GCAL] Perform requests concurrently

fmartingr:
#### Summary

The google calendar provider can't make batch request properly since the API does not support it so this PR adds a separate way to request from the Google Calendar API by using workers to do requests in parallel.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53896


server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+
+	ctxTimeout, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	for i := 0; i <= concurrency; i++ {
+		go func(m mscalendar, c context.Context, w *sync.WaitGroup, in chan store.User, out chan StatusSyncJobSummary) {
+			for {
+				select {
+				case <-c.Done():
+					m.Logger.Errorf("Tiemout processing users availability")
```
mickmister, line 155:
```suggestion
					m.Logger.Errorf("Timeout processing users availability")
```

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
-		var err error
-		calendarViews, err = m.GetCalendarViews(users)
-		if err != nil {
-			return users, calendarViews, errors.Wrap(err, "not able to get calendar views for connected users")
-		}
+	go func() {
+		wg.Wait()
+		close(in)
+		// Ensure information was pulled successfully
+		time.Sleep(250 * time.Millisecond)
```
mickmister, line 227: Why is 250ms chosen specifically?

mickmister (approved):
LGTM, though asking @hanzei for a review on the goroutine orchestration

server/mscalendar/availability.go
fmartingr: No specific reason, just trying to check what was the problem with the tests, but also leaving some room for the `out` channel to be consumed before closing it. We can probably work with a lower value.

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+	}
+
+	return users, calendarViews, nil
+}
+
+func (m *mscalendar) retrieveUsersToSyncUsingGoroutines(ctx context.Context, userIndex store.UserIndex, syncJobSummary *StatusSyncJobSummary, concurrency int) ([]*store.User, []*remote.ViewCalendarResponse, error) {
+	start := time.Now().UTC()
+	end := time.Now().UTC().Add(calendarViewTimeWindowSize)
+
+	in := make(chan store.User, concurrency*4)
```
hanzei, line 143: Why does the job queue need to be buffered?

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+	out := make(chan StatusSyncJobSummary, concurrency*4)
```
hanzei, line 144: Same question here.

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+		if !(user.Settings.UpdateStatus || user.Settings.ReceiveReminders) {
+			continue
+		}
+
+		wg.Add(1)
+		in <- *user
+
+		users = append(users, user)
 	}
+
```
hanzei, line 216: Can we `close(in)` here already?

server/mscalendar/availability.go
hanzei: The code here does seam right. If `wg.Wait()` returned, all workers should have returned. Why does the code need to wait before we close `out`?

server/mscalendar/availability.go
```diff
@@ -27,6 +29,9 @@ const (
 	upcomingEventNotificationWindow = (StatusSyncJobInterval * 11) / 10 // 110% of the interval
 	logTruncateMsg                  = "We've truncated the logs due to too many messages"
 	logTruncateLimit                = 5
+
+	// Default concurrency to use for calendar providers that doesn't allow batch requests
+	defaultConcurrency = 4
```
hanzei, line 34: The idiomatic way to start a godoc comment is with the variable name i.e. `defaultConcurrency is [....]`

hanzei (changes requested):
Surely these changes will make the the sync process more performant. :+1: 

There are still some rough edges with the code, see my comments below.

hanzei:
@fmartingr https://itnext.io/explain-to-me-go-concurrency-worker-pool-pattern-like-im-five-e5f1be71e2b0 might be a good read, as it's very similar to what you are implementing here.

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
-				m.Logger.Warnf("Not able to enable active user %s from user index. err=%v", user.MattermostUserID, err)
-				continue
+	var err error
+	calendarViews, err = m.GetCalendarViews(users)
+	if err != nil {
+		return users, calendarViews, errors.Wrap(err, "not able to get calendar views for connected users")
+	}
+
+	if len(calendarViews) == 0 {
+		return users, calendarViews, fmt.Errorf("no calendar views found")
```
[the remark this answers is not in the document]
fmartingr: Shouldn't we move away from `github.com/pkg/errors` since it's not maintained anymore?

server/mscalendar/availability.go
```diff
@@ -115,37 +118,119 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
 		}
 
 		users = append(users, user)
+	}
+	if len(users) == 0 {
+		return users, calendarViews, errNoUsersNeedToBeSynced
```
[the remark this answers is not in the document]
fmartingr: I agree but I wanted to maintain as much original logic from the plugin intact, to reduce the number of regressions we may have when everything in merged into the same branch.

fmartingr:
Hey @hanzei, thanks a lot for the review. It's been a while since I wrote some concurrent code and it seems that I forgot the bare basics, I think yesterday I was aiming to consume both channels before closing them for whatever reason, I've fixed that. Also, instead of using the same loop to add items to the `input` channel, I move that to a separate goro which is easier to follow, since the other loop has some other logic inside that would be troublesome to work with. Does it look better now?

server/mscalendar/availability.go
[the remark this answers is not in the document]
hanzei: `github.com/pkg/errors` is stable and widely used (in Mattermost and outside of it). 2/5 to stick with it for now.

If we wanted to migrate away, then we should do that in a separate effort. #consitency

server/mscalendar/availability.go
```diff
@@ -155,14 +134,135 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+
+			syncJobSummary.NumberOfUsersFailedStatusChanged += js.NumberOfUsersFailedStatusChanged
+			calendarViews = append(calendarViews, js.CalendarEvents)
+		case <-finishChan:
+			if len(calendarViews) == 0 {
+				return users, calendarViews, fmt.Errorf("no calendar views found")
+			}
+
+			return users, calendarViews, nil
+		}
```
hanzei, line 245: Why is this case needed? Doesn't closing `out` already indicate that there isn't more to read?

server/mscalendar/availability_test.go
```diff
@@ -399,11 +403,13 @@ func TestRetrieveUsersToSyncIndividually(t *testing.T) {
 		m := New(env, "").(*mscalendar)
 		jobSummary := &StatusSyncJobSummary{}
 
-		_, _, err := m.retrieveUsersToSync([]*store.UserShort{}, jobSummary, true)
+		_, _, err := m.retrieveUsersToSyncUsingGoroutines(ctx, []*store.UserShort{}, jobSummary, concurrency)
 		require.ErrorIs(t, errNoUsersNeedToBeSynced, err)
 	})
 
 	t.Run("user reminders and status disabled", func(t *testing.T) {
+		ctx := context.TODO()
```
hanzei, line 411: Nit: Should there be a test that checks if cancelling `ctx` returns an error? If we have that test, this can become a `context.Background()`.

server/mscalendar/availability.go
```diff
@@ -27,6 +29,9 @@ const (
 	upcomingEventNotificationWindow = (StatusSyncJobInterval * 11) / 10 // 110% of the interval
 	logTruncateMsg                  = "We've truncated the logs due to too many messages"
 	logTruncateLimit                = 5
+
+	// defaultConcurrency is the default number of workers to span for calendar providers that doesn't allow batch requests
+	defaultConcurrency = 4
```
hanzei, line 34: Why was this value picked? What are the tradeoffs?

server/mscalendar/availability.go
```diff
@@ -155,14 +134,135 @@ func (m *mscalendar) retrieveUsersToSync(userIndex store.UserIndex, syncJobSumma
...
+			continue
+		}
+
+		// If user does not have the proper features enabled, just go to the next one
+		if !(user.Settings.UpdateStatus || user.Settings.ReceiveReminders) {
+			continue
+		}
+
+		users = append(users, user)
+	}
```
hanzei, line 163: Not blocking: What I liked about the version was that the workers already start while more users are getting loaded. It's not a big deal, but why should the works wait until all users have been loaded?

hanzei (changes requested):

hanzei:
That looks much better @fmartingr :+1: There are still some more small comments from my side.

server/mscalendar/availability.go
fmartingr: I have zero insights into customers environments, but that seemed a reasonable number to have to achieve proper concurrency/parallelism if running server has at least 4c/t. Using a higher number can give better results but also diminishing returns if the server can't handle all at once, and I don't want to hammer a lot of goros to the scheduler to just let them wait for each other. If this where a CLI tool or something I would use `runtime.NumCPU()` but I don't want the plugin to use all available resources every 5 minutes, even if most of the time it's going to be waiting for responses. What do you think? Is there any other plugin doing something similar?

server/mscalendar/availability.go
fmartingr: it was pure cosmetic preference, didn't want to put all the logic in the same `case` statement, but it makes sense.

server/mscalendar/availability.go
fmartingr: - I wanted to spawn the workers only if there was actually work to do (`len(users) > 0`)
- This loop would need to be in a goro as well, and it has some other data transfer that need to be taken care of (`users`, `numberOfLogs`, `syncJobSummary`). It would make this function more complex.

server/mscalendar/availability.go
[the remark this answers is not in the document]
fmartingr: I don't think it makes a difference here, but 5/5 on #consistency.

server/mscalendar/availability.go
hanzei: >  don't want the plugin to use all available resources every 5 minutes, even if most of the time it's going to be waiting for responses.

I don't think that's what is happening here. As you pointed out, the workers will mostly wait for an HTTP response. During that time the go runtime is smart enough to pass control of the CPU to other processes.

 >  Is there any other plugin doing something similar?
 
 Not on that scale. E.g. the GitHub plugin does concurrent API requests, but only four.
 
 Do you know if the MS calendar API punishes a lot of requests coming from the same IP? I.e. can we use an aggressive number here?

fmartingr: > Do you know if the MS calendar API punishes a lot of requests coming from the same IP? I.e. can we use an aggressive number here?

Yes, that was the other reason I forgot about yesterday. Google APIs have per project and per user quotas. Though the documentation says that its for write requests, I don't know if we can trigger some limit if we do a lot of reads too.

server/mscalendar/availability_test.go
fmartingr: What would be a correct way of testing a context being cancelled? I've tried some stuff but since this runs in goros results with gomock are not consistent, and I'm not sure if using `.AnyTimes()` is a good solution here.

hanzei: You can pass a cancelled context to `retrieveUsersToSyncUsingGoroutines`. This is not the same as waiting for the timeout, but good enough.

server/mscalendar/availability.go
fmartingr: Yes, but we are building the gcal plugin on top of the mscalendar, by using different remotes when retrieving data, but maintaining the plugin UI/UX.

hanzei: In that case, I would suggest moving the concurrency variable out of the function (maybe into `mscalendar`?) and configure it per backend.

fmartingr: I can do that as well, though the mscalendar allows for batch requests so it's never going to reach this piece of code.

hanzei:
@fmartingr Please let me know when the PR is ready for a final round of review

PR #309, opened 2023-08-24
[GCAL/MSCAL] Create event UX improvements

fmartingr:
#### Summary
- Added "Location" and "Link to channel" optional field text
- Time pickers now have defaults depending on the selected date
- Custom icon in "Create event" menu item
  Screenshot 2023-08-25 at 13 43 56
- Added custom icon to date selector:
  Screenshot 2023-08-25 at 13 56 35

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53998


webapp/src/components/modals/create_event_form.css
```diff
@@ -0,0 +1,6 @@
+.create-event-form input[type="date"]::-webkit-calendar-picker-indicator {
```
mickmister, line 1: Can we use a more specific class name, like `mscalendar-create-event-form`? This way it's definitely scoped to this plugin.

webapp/src/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
mickmister, line 1: :+1: A pain that this is required though

webapp/tsconfig.json
```diff
@@ -25,6 +27,9 @@
     "experimentalDecorators": true,
     "jsx": "react"
   },
+  "files": [
+    "src/custom.d.ts"
+  ],
```
mickmister, line 32: I don't think I've ever used `files` like this for `.d.ts` files. Maybe we can move `custom.d.ts` into `types/svg.d.ts`, and then set tsconfig's `typeRoots` to `./types`. I think that's the proper way to do this.

But if it ain't broke, don't fix it? 🤷

webapp/src/index.tsx
```diff
@@ -27,7 +30,7 @@ export default class Plugin {
             await store.dispatch(getProviderConfiguration());
 
             registry.registerChannelHeaderMenuAction(
-                <span><i className='icon fa fa-calendar-plus-o'/>{'Create calendar event'}</span>,
+                <span><Icon image={CalendarOutlineIcon}/>{'Create calendar event'}</span>,
```
mickmister, line 33: This line is getting kind of long

```suggestion
                (
                    <span>
                        <Icon image={CalendarOutlineIcon}/>
                        {'Create calendar event'}
                    </span>
                ),
```

0/5 on the parens that I added there, but definitely think it should be multiple lines

webapp/src/components/icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+            <i
+                className='icon fa'
+                style={{
+                    width: '18px',
+                    height: '18px',
+                    backgroundImage: `url(${this.props.image})`,
+                    backgroundPosition: '50% 50%',
+                    backgroundRepeat: 'no-repeat',
+                    verticalAlign: 'middle',
+                }}
```
mickmister, line 27: Is this taken from somewhere else? It seems very specific (size etc.) for a generic-sounding `Icon` component. It seems fine though.

webapp/src/components/modals/create_event_form.css
mickmister: Also we generally prefer `scss` over `css` files, which allow you to do things like nesting:

```suggestion
.create-event-form {
    input[type="date"]::-webkit-calendar-picker-indicator {
        background: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M12 12H17V17H12V12M19 3H18V1H16V3H8V1H6V3H5C3.9 3 3 3.9 3 5V19C3 20.1 3.9 21 5 21H19C20.1 21 21 20.1 21 19V5C21 3.9 20.1 3 19 3M19 5V7H5V5H19M5 19V9H19V19H5Z' /%3E%3C/svg%3E") no-repeat;
        display: block;
        height: 18px;
        width: 18px;
    }
}
```

mickmister (approved):
LGTM, a few non-blocking suggestions

webapp/src/custom.d.ts
fmartingr: You tell me. 😅

webapp/src/components/icon.tsx
fmartingr: I just played with the styles until it seemed right.

webapp/tsconfig.json
fmartingr: Let's do it the proper way.

webapp/src/components/modals/create_event_form.css
mickmister: Thoughts about using the class name `mscalendar-create-event-form` instead? We usually try to scope css classes to a given plugin whenever possible

mickmister (approved):

webapp/src/components/modals/create_event_form.css
fmartingr: Oh yeah I got lost in thought about using `mscalendar` as class name and I forgot to put it while making other changes.

PR #314, opened 2023-08-31
[GCAL/MSCAL] Use `npm run bulid` when making plugin bundles

fmartingr:
#### Summary

Changed the build command for the webapp to `build` so it can be deployed in production enviornments.


mickmister (approved):
I think this was a change I introduced by mistake. Should not have been committed. Thanks @fmartingr :+1:

PR #374, opened 2024-06-26
Clarify error message when calendar account is connected to disabled user

hanzei:
#### Summary
If a disabled user is connected, any mm user that wants to connect to the same calendar account will get an error like `XYZ is already mapped to a Mattermost account, but the Mattermost user could not be found`. This PR changes the error message to make it more clear that the user is connected to a disabled account.

#### Ticket Link
https://hub.mattermost.com/private-core/pl/y9nciw3nrjbg9jwf3mjtk1asbo

### DIFF
--- a/calendar/engine/oauth2.go
+++ b/calendar/engine/oauth2.go
@@ -19,8 +19,11 @@ import (
 
 const BotWelcomeMessage = "Bot user connected to account %s."
 
-const RemoteUserAlreadyConnected = "%s account `%s` is already mapped to Mattermost account `%s`. Please run `/%s disconnect`, while logged in as the Mattermost account"
-const RemoteUserAlreadyConnectedNotFound = "%s account `%s` is already mapped to a Mattermost account, but the Mattermost user could not be found"
+const (
+	RemoteUserAlreadyConnected         = "%s account `%s` is already mapped to Mattermost account `%s`. Please run `/%s disconnect`, while logged in as the Mattermost account"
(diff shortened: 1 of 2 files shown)


calendar/engine/oauth2.go
```diff
@@ -19,8 +19,11 @@ import (
 
 const BotWelcomeMessage = "Bot user connected to account %s."
 
-const RemoteUserAlreadyConnected = "%s account `%s` is already mapped to Mattermost account `%s`. Please run `/%s disconnect`, while logged in as the Mattermost account"
-const RemoteUserAlreadyConnectedNotFound = "%s account `%s` is already mapped to a Mattermost account, but the Mattermost user could not be found"
+const (
+	RemoteUserAlreadyConnected         = "%s account `%s` is already mapped to Mattermost account `%s`. Please run `/%s disconnect`, while logged in as the Mattermost account"
+	RemoteUserAlreadyConnecteDisabled  = "%s account `%s` is already mapped to a Mattermost account, but the account is deactived. Please enable it and run `/%s disconnect`,  while logged in as the other Mattermost account, and try again"
```
ayusht2810, line 24:
```suggestion
	RemoteUserAlreadyConnectedDisabled  = "%s account `%s` is already mapped to a Mattermost account, but the account is deactivated. Please enable it and run `/%s disconnect`,  while logged in as the other Mattermost account, and try again"
```

ayusht2810 (approved):

mickmister (approved):

calendar/engine/oauth2.go
hanzei: Good catch :+1:

PR #377, opened 2024-07-05
fix: reduce error logging verbosity

fmartingr:
#### Summary
Reduce logging verbosity by lowering the level of the main offender to `warn` from `error`, also adds more information in error messages to be able to track those properly.


wiggin77 (approved):
LGTM :+1:

PR #379, opened 2024-07-22
Add API to list calendars for a user

ayusht2810:
#### Summary
- Added API to list calendars
- Updated create event  API to use calendarID while creating an event

#### Ticket Link
Related to https://github.com/mattermost/mattermost-plugin-google-calendar/issues/59

### DIFF
--- a/calendar/api/events.go
+++ b/calendar/api/events.go
@@ -36,6 +36,7 @@ type createEventPayload struct {
 	Subject     string `json:"subject"`
 	Location    string `json:"location,omitempty"`
 	ChannelID   string `json:"channel_id"`
+	CalendarID string `json:"calendar_id"`
 }
 
 func (cep createEventPayload) ToRemoteEvent(loc *time.Location) (*remote.Event, error) {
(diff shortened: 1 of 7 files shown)


calendar/config/const.go
```diff
@@ -27,6 +27,8 @@ const (
 
 	InternalAPIPath   = "/api/v1"
 	PathEvents        = "/events"
+	PathCalendar      = "/calendar"
+	PathList          = "/list"
```
raghavaggarwal2308, line 31:
```suggestion
	PathListCalendar          = "/calendar/list"
```

calendar/api/events.go
```diff
@@ -272,3 +274,31 @@ func (api *api) createEvent(w http.ResponseWriter, r *http.Request) {
 
 	httputils.WriteJSONResponse(w, `{"ok": true}`, http.StatusCreated)
 }
+
+func (api *api) listCalendars(w http.ResponseWriter, r *http.Request) {
+	mattermostUserID := r.Header.Get(HeaderMattermostUserID)
+	if mattermostUserID == "" {
+		httputils.WriteUnauthorizedError(w, fmt.Errorf("unauthorized"))
```
raghavaggarwal2308, line 281: Don't we need to log the error here?

calendar/config/const.go
ayusht2810: discussed

wiggin77 (approved):
LGTM :+1:

wiggin77 (changes requested):
I agree with copilot.  It appears a calendar ID field has been added but is hard-coded to an empty string.  I'm not understanding how that addresses the underlying issue reported.

PR #380, opened 2024-07-23
fix: avoid using output from failing function

fmartingr:
#### Summary
This pull request fixes a crash that was caused when processing the daily summary for an user that couldn't be retrieved. Since we use the output from the `LoadUser` call in the `Warnf` call, plugin was crashing.


mgdelacroix (approved):
LGTM! :+1:

wiggin77 (approved):

PR #397, opened 2024-08-14
[GH-394] Add timeout to http client

### LINKED ISSUE #394: HTTP client without timeout
An HTTP client is configured without specifying a timeout duration. This makes it so any outgoing HTTP requests with this client may hang indefinitely and never time out.

msgraph/remote.go
```go
httpClient := &http.Client{}
c := &client{
	conf:       r.conf,
	ctx:        ctx,
	httpClient: httpClient,
	Logger:     r.logger,
	rbuilder:   msgraph.NewClient(httpClient),
}
```

It’s recommended to configure the client with a reasonable timeout value for its use-case.

ex. `httpClient := &http.Client{Timeout: time.Second * 5}`

See https://mattermost.atlassian.net/browse/MM-60185

raghavaggarwal2308:
#### Summary
- Add timeout to http client
- Http client has a [default timeout](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/httpclient/httpclient-timeout-method#remarks) of 100 sec. We updated it to ~5~ 60 sec.

#### Ticket Link
Fixes #394 
Fixes https://mattermost.atlassian.net/browse/MM-60185

### DIFF
--- a/msgraph/remote.go
+++ b/msgraph/remote.go
@@ -53,7 +53,7 @@ func (r *impl) MakeClient(ctx context.Context, token *oauth2.Token) remote.Clien
 // MakeSuperuserClient creates a new client used for app-only permissions.
 func (r *impl) MakeSuperuserClient(ctx context.Context) (remote.Client, error) {
 	httpClient := &http.Client{
-		Timeout: time.Second * 5,
+		Timeout: time.Second * 60,
 	}
 	c := &client{
 		conf:       r.conf,
--- a/msgraph/remote.go
+++ b/msgraph/remote.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/microsoft"
@@ -51,7 +52,9 @@ func (r *impl) MakeClient(ctx context.Context, token *oauth2.Token) remote.Clien
 
 // MakeSuperuserClient creates a new client used for app-only permissions.
 func (r *impl) MakeSuperuserClient(ctx context.Context) (remote.Client, error) {
-	httpClient := &http.Client{}
+	httpClient := &http.Client{
+		Timeout: time.Second * 5,
+	}
 	c := &client{
 		conf:       r.conf,
 		ctx:        ctx,


msgraph/remote.go
```diff
@@ -51,7 +52,9 @@ func (r *impl) MakeClient(ctx context.Context, token *oauth2.Token) remote.Clien
 
 // MakeSuperuserClient creates a new client used for app-only permissions.
 func (r *impl) MakeSuperuserClient(ctx context.Context) (remote.Client, error) {
-	httpClient := &http.Client{}
+	httpClient := &http.Client{
+		Timeout: time.Second * 5,
```
wiggin77, line 56:
```suggestion
		Timeout: time.Second * 60,
```

wiggin77 (approved):

msgraph/remote.go
```diff
@@ -51,7 +52,9 @@ func (r *impl) MakeClient(ctx context.Context, token *oauth2.Token) remote.Clien
 
 // MakeSuperuserClient creates a new client used for app-only permissions.
 func (r *impl) MakeSuperuserClient(ctx context.Context) (remote.Client, error) {
-	httpClient := &http.Client{}
+	httpClient := &http.Client{
+		Timeout: time.Second * 60,
```
enzowritescode, line 56: 60 seconds seems like a lot. What was the rationale here?

raghavaggarwal2308: @wiggin77 Can you please look into the above?

enzowritescode: I see in your description for the PR you reference https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/httpclient/httpclient-timeout-method#remarks but that seems irrelevant here.

The default timeout for https://pkg.go.dev/net/http#Client is actually 0 (which means no timeout)

```
package main

import (
	"fmt"
	"net/http"
)

func main() {
	client := http.Client{}

	fmt.Println(client.Timeout)
}
```

Output:
```
0s
```

enzowritescode: If we don't have any data to help our decision, 10s would probably be safe.

wiggin77: 5s in my experience is too short and leads to spurious errors.  I chose 60 seconds as balance between handling spurious interruptions and not piling up connections.  This is not a section of code that will generate a ton of concurrent connections.

enzowritescode (approved):

Issue #392, opened 2024-08-13
The help text for viewcal slash command does not matches the response


AayushChaudhary0001:
While testing the plugin, I found that the help text for viewcal slash command does not match with the response of the slash command. The helper text shows that the slash command will show the events for the upcoming week, whereas, the response shows all the events for the ongoing week without considering the current day.

**Steps to reproduce:-**
1. Connect the Microsoft account with the Mattermost account.
2. Run the "/mscalendar viewcal" slash command.

The helper text shows that the command will show the events in the upcoming week, whereas the response shows all the events in the ongoing week(for previous days as well).

**Lets take an example here:-** 
I ran the viewcal slash command here on Tuesday, but the response consisted of the events which were already planned and completed on Monday(a day before). Also, as the helper text shows the upcoming week, the current week should not be considered.

![image](https://github.com/user-attachments/assets/2bc56c48-8995-4719-a76a-8d5ceaa6cff8)

![image](https://github.com/user-attachments/assets/3e4f93d9-f17c-443a-8e26-70ba078fbbd4)

**Suggestion:-** If we change the word "upcoming" to "ongoing/current", then every case will be handled and the situation would be solved.

raghavaggarwal2308:
@wiggin77 Can someone help us in understanding about what exactly is the expected behaviour of this command?

wiggin77:
@esethna would you prefer the wording changed or the behaviour?  See above.

esethna:
@wiggin77 @raghavaggarwal2308 I'd propose we change the help text to `View your events for the week.`

If it's easy to remove the tables for days in the past that would be a good win, but not necessary if it adds too much complexity.

Kshitij-Katiyar:
@wiggin77 @esethna @AayushChaudhary0001 

The following function retrieves events when the /viewcal command is triggered. This command fetches all events scheduled between `current time - 24 hours` and `current time + 14 * 24 hours`.
```
events, err := c.Engine.ViewCalendar(c.user(), time.Now().Add(-24*time.Hour), time.Now().Add(14*24*time.Hour))
```

![Screenshot from 2024-08-14 10-46-46](https://github.com/user-attachments/assets/046a54f7-3e2d-4d29-9cce-7913b4d41b79)
![Screenshot from 2024-08-14 10-48-48](https://github.com/user-attachments/assets/9e069acd-f8e3-4d2d-89dd-c89b79e68a48)
[...]

PR #399, opened 2024-08-14
[MM-392]: Fixed the issue of viewcal getting events of the previous day

Kshitij-Katiyar:
#### Summary
- Fixed the issue of `/mscalendar viewcal` command getting events of the previous day also
- Updated the help text of the `/mscalendar viewcal` to be more meaningful

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-mscalendar/issues/392

##### Existing
Screenshot from 2024-08-14 14-18-10
##### Updated
Screenshot from 2024-08-14 14-16-51


calendar/command/command.go
```diff
@@ -54,7 +54,7 @@ var cmds = []*model.AutocompleteData{
 			model.NewAutocompleteData("disable", "", "Disable your daily summary."),
 		},
 	},
-	model.NewAutocompleteData("viewcal", "", "View your events for the upcoming week."),
+	model.NewAutocompleteData("viewcal", "", "View your events for the current and upcoming week."),
```
wiggin77, line 57: @Kshitij-Katiyar  If I ran this command on a Wednesday, would it show events for the Wednesday, Thursday, Friday, and every day the week after?

raghavaggarwal2308: @wiggin77 Please refer to this comment for more clarity: https://github.com/mattermost/mattermost-plugin-mscalendar/issues/392#issuecomment-2293011803

wiggin77 (approved):

AayushChaudhary0001:
@Kshitij-Katiyar Can we change the helper text here?
The current logic says that the events will be fetched for current and upcoming weeks, but if we run this command mid-week, then in general it will be considered starting Sunday/Monday which will not match the logic behind the functionality.
Instead, we can use something which fetches the events for next 14 days/2 weeks starting from today in helper text.

Kshitij-Katiyar:
> @Kshitij-Katiyar Can we change the helper text here? The current logic says that the events will be fetched for current and upcoming weeks, but if we run this command mid-week, then in general it will be considered starting Sunday/Monday which will not match the logic behind the functionality. Instead, we can use something which fetches the events for next 14 days/2 weeks starting from today in helper text.

Hey @AayushChaudhary0001, I have updated the help text to `View your events for the upcoming 14 days, including today.`
let me know if you think this is better.

AayushChaudhary0001 (approved):
Tested this PR, new helper text looks good, Approved.

raghavaggarwal2308:
@wiggin77 Can you please give this PR another review for changes made by suggestions from by @AayushChaudhary0001

wiggin77:
> if we run this command mid-week, then in general it will be considered starting Sunday/Monday
 
@AayushChaudhary0001 previously when I asked what would happen if the command was run on a Wednesday, the answer was it would include Wed, Thurs, Fri, and then each day of the next week. The statement above seems to contradict that.

> I have updated the help text to `View your events for the upcoming 14 days, including today.`

@Kshitij-Katiyar I cannot approve this until I know exactly what the output will be when running the command on any given day.  IMO, displaying the current day, plus remaining days in current week, plus the following week, was correct.  We can live with just displaying the next 14 days, but I need confirmation that is indeed the behaviour no matter when the command is run.

raghavaggarwal2308:
@wiggin77 It will display the events from current day (When a user runs the command) and the events from 14 days after that. I mentioned the same in my comment [here](https://github.com/mattermost/mattermost-plugin-mscalendar/issues/392#issuecomment-2293011803). 
Please let me know if there is some confusion.

wiggin77 (approved):
LGTM :+1:

PR #404, opened 2024-08-22
[MM-60217]: Converted the reserved Markdown characters to their HTML entity equivalents

Kshitij-Katiyar:
#### Summary
Converted the reserved Markdown characters to their HTML entity equivalents. So no markdown can be rendered


wiggin77 (changes requested):
Thanks @Kshitij-Katiyar .   I will always ask for unit tests for new code like this.

wiggin77:
@Kshitij-Katiyar see comment here:  https://mattermost.atlassian.net/browse/MM-60217?focusedCommentId=169185

enzowritescode:
I am still seeing rendered markdown when events are updated

md_link

enzowritescode (commented):
I am still seeing rendered markdown when events are updated

Kshitij-Katiyar:
@enzowritescode fixed it.

enzowritescode:
update

This does not look correct

Kshitij-Katiyar:
> update
> 
> This is not getting displayed correctly

@enzowritescode Can you share whats the event name is now after the updation ?

enzowritescode:
The event is titled "[DuckDuckGo](https://duckduckgo.com/)" (but without the quotes)

raghavaggarwal2308:
@enzowritescode @wiggin77 

> The event is titled `[DuckDuckGo](https://duckduckgo.com/)`

We are unable to render any html code at the end of URL on Mattermost. I think is is being considered as a query param instead of a html code. Look at the screenshots for more clarity. Do you have any suggestions on how we can fix it?

image
image
image

enzowritescode:
The front-end must be performing some processing on things that it believes to be markdown which we don't want. I will look into this some more.

```
<a class="theme markdown__link" href="https://duckduckgo.com/&amp;#41" rel="noreferrer" target="_blank">https://duckduckgo.com/&amp;#41</a>
```

wiggin77 (approved):

enzowritescode (approved):

enzowritescode:
@Kshitij-Katiyar escaping the square brackets and parenthesis wasn't enough. I added forward slash and period so plain URLs wouldn't end up getting treated as links. I also put the the `replacements` map in numerical order so it was a little easier to read.

AayushChaudhary0001:
I was testing this PR and found an issue in this PR.
Text if entered in strikethrough format in MS Calendar is still reflected as strikethrough text in Mattermost, the rest of everything works fine, but this strikethrough needs to be fixed.

I am attaching a screenshot below:-

image

raghavaggarwal2308:
@AayushChaudhary0001 The above issue should be fixed now, please re-test

AayushChaudhary0001 (approved):
Tested this PR after the fix, strikethrough text is displayed fine now. Apart from this, every other markdown has not been impacted due to the lastest fix. Everything works fine, LGTM. Approved.

PR #420, opened 2024-10-16
[MM-872]: Added server testcases for api/post_action.go

Kshitij-Katiyar:
#### Summary
Added server testcases for api/post_action.go

### DIFF
--- a/calendar/api/post_action_test.go
+++ b/calendar/api/post_action_test.go
@@ -1,0 +1,619 @@
+package api
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/config"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/engine"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/remote"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/store"
+
+	"github.com/mattermost/mattermost/server/public/model"
+)
+
+func TestPreprocessAction(t *testing.T) {
+	api, _, _, _, _, _, _, _ := GetMockSetup(t)
+
+	tests := []struct {
+		name       string
+		setup      func(*http.Request)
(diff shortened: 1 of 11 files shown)


wiggin77 (approved):
LGTM :+1:

calendar/api/post_action_test.go
```diff
@@ -0,0 +1,619 @@
...
+			},
+		},
+		{
+			name: "Invalid request body",
+			setup: func(req *http.Request) {
+				req.Header.Set(MMUserIDHeader, MockUserID)
+				req.Body = io.NopCloser(bytes.NewBufferString("invalid json"))
+			},
+			assertions: func(rec *httptest.ResponseRecorder, mscal engine.Engine, user *engine.User, eventID, option, postID string) {
+				assert.Equal(t, http.StatusOK, rec.Result().StatusCode)
```
raghavaggarwal2308, line 51: Why are we getting status OK in all the cases. Please check if we need to update our code to return proper status codes

calendar/api/post_action_test.go
```diff
@@ -0,0 +1,619 @@
...
+					Context: map[string]interface{}{
+						"value":     true,
+						"change_to": "away",
+					},
+				}
+				bodyBytes, _ := json.Marshal(requestBody)
+				req.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+			},
+			assertions: func(rec *httptest.ResponseRecorder) {
+				assert.Equal(t, http.StatusOK, rec.Result().StatusCode)
```
raghavaggarwal2308, line 539: Same here

calendar/api/test_utils.go
```diff
@@ -39,6 +41,64 @@ func (m *MockNotificationProcessor) Enqueue(notifications ...*remote.Notificatio
 func (m *MockNotificationProcessor) Configure(env engine.Env) {}
 func (m *MockNotificationProcessor) Quit()                    {}
 
+type mockClient struct{}
```
raghavaggarwal2308, line 44: Can't we use mockgen for this? OR this coming from an external library?

raghavaggarwal2308 (changes requested):

calendar/api/post_action_test.go
Kshitij-Katiyar: We are not writing any status code in the whole file. I think we should make it a separate task to return proper status code for all the API in the plugin

raghavaggarwal2308: @Kshitij-Katiyar Sure, please log a issue for the same

raghavaggarwal2308 (approved):

PR #429, opened 2024-11-05
[MM-844]: Added server testcase for store/welcome_store.go

Kshitij-Katiyar:
### Summary
Added server testcase for store/welcome_store.go


calendar/store/welcome_store_test.go
```diff
@@ -0,0 +1,141 @@
...
+			assertions: func(t *testing.T, resp string, err error) {
+				require.Equal(t, "", resp)
+				require.ErrorContains(t, err, "failed plugin KVdelete", "KVDelete failed")
+			},
+		},
+		{
+			name: "Success deleting user welcome post",
+			setup: func(mockAPI *testutil.MockPluginAPI) {
+				mockAPI.On("KVGet", mock.AnythingOfType("string")).Return([]byte(`"mockPostID"`), nil)
+				mockAPI.On("KVDelete", mock.Anything, mock.Anything).Return(nil)
```
raghavaggarwal2308, line 121: Can we replace mock.Anything with AnythingOfType?

raghavaggarwal2308 (approved):

wiggin77 (changes requested):
CI errors to fix.

calendar/store/welcome_store_test.go
wiggin77: In general to keep this more readable, you can create vars for the major mock types you use. e.g. 
```go
mockString := mock.AnythingOfType("string")
```

then 

```go
mockAPI.On("KVGet", mockString).Return([]byte(`"mockPostID"`), nil)
```

Also, where does KVDelete take two args?

Kshitij-Katiyar: @wiggin77  Fixed the CI & Replaced `mock.Anything` with `mock.AnythingOfType`

calendar/store/welcome_store_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	tests := []struct {
+		name       string
+		setup      func(*testutil.MockPluginAPI)
+		assertions func(*testing.T, error)
+	}{
+		{
+			name: "Error storing user welcome post",
+			setup: func(mockAPI *testutil.MockPluginAPI) {
+				mockAPI.On("KVSet", MockString, mock.Anything).Return(&model.AppError{Message: "KVSet failed"})
```
wiggin77, line 68: Let's completely remove all 'mock.Anything' and only use typesafe args.

wiggin77 (changes requested):

calendar/store/welcome_store_test.go
Kshitij-Katiyar: @wiggin77 
Value of type []byte were not allowed to be mocked using `mock.AnythingOfType` so used `mock.MatchedBy` to keep the testcases typesafe
```
mock.MatchedBy(func(arg interface{}) bool {
        _, ok := arg.([]byte)
        return ok
    }),
```

calendar/store/test_util.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/testutil"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/tracker/mock_tracker"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/utils/bot/mock_bot"
+)
+
+var MockString = mock.AnythingOfType("string")
```
wiggin77, line 14: Can you put the matched by here as well.  Eventually we will move these into a util file and re-use.

wiggin77 (changes requested):

wiggin77 (approved):
LGTM :+1:

PR #486, opened 2026-04-03
MM-68138 Adding experimental calendar UI sidebar

avasconcelos114:
#### Summary

This PR adds two main things:
1. The missing implementation for creating calendar events for MS Calendar, which can be used via `/mscalendar event create` (which had been previously hidden in MS Cal)
2. A calendar UI that is shown when the `Enable Experimental Calendar UI` setting is enabled, the UI features:
  - A right-hand sidebar displaying daily and weekly views. The view switches automatically based on sidebar width (daily below 500px, weekly above). Includes a now-indicator, all-day slot, and 30-minute time slots.
  - Event tooltips that appear when clicking an event, which feature some basic information on it, and a join meeting button when a meeting has been detected 
  - Clicking on a time slot in the calendar opens the create calendar event modal with the time slot pre-selected onto the modal
  - Periodic polling and refreshing mechanism to keep the calendar events up-to-date
  - Detection of account connection state (calendar only visible when a connected account is available)
  - Theme aware and should be clearly visible on both light and dark themes
  - Takes user's timezone settings into consideration when displaying the calendar event times

#### Screenshots

|  |  |
| :--: | :--: |
| image | image |
#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-68138

#### QA Steps
As this is a whole set of new features, there are multiple things that will require testing for this:

**Calendar Sidebar**
* The "Enable Experimental Calendar UI" setting adds the appbar button to toggle the sidebar
* Having a disconnected account correctly shows the sidebar in its unauthenticated state
* Completing the auth flow automatically changes the sidebar to authenticated and the calendar is displayed
* The theme of the calendar is consistent with the active theme of the platform (only kicks in when fully applied due to plugin constrains)
* The timezone selected in the user settings correctly accounts for what the time of the meeting is shown as
* When in extended view, the calendar switches to a weekly view, and a daily view is shown when the sidebar is compressed
* Clicking on an event displays its information correctly (event title, creator, RSVP status, online meeting link)
* Clicking on an empty slot in the calendar triggers the event creation modal (further specs below), and creating an event will immediately update the events in the sidebar
* When events are created outside of the app, the sidebar will sync the newly created event within the next IRL minute (60sec intervals)
  * OR you can press the refresh button on top of the sidebar to immediately have it synced

**Event Create Modal**
* The modal can be triggered via the `/mscalendar event create`, which is no longer hidden from autocomplete
* When triggered via the sidebar, the date, start/end times for the event have been pre-filled to the time slot that was clicked
* Like the Google Calendar plugin, you can add guests and link the event to a channel
* The description field will be mapped to the description of the event when seen in the outlook calendar 

## Change Impact: 🟡 Medium

**Reasoning:** The PR adds a large, feature‑gated experimental calendar UI across backend and frontend, and it also ungates the create-event slash command—changes affect both server routes and the webapp surface area, increasing scope and integration points while remaining feature-flagged.

**Regression Risk:** Moderate — risk comes from new API surface (the /view endpoint and provider configuration adjustments), timezone normalization changes, request/cancellation logic, and many new frontend modules (reducers, hooks, components) that can introduce UI or state regressions. Some backend and API changes are covered by new tests, but the expanded webapp codebase has limited end-to-end coverage relative to its size.

**QA Recommendation:** Medium-intensity manual QA recommended. Prioritize: slash-command behavior and connection gating; create-event end-to-end flows including timezone handling and error paths; sidebar UX (daily/weekly switching, tooltips, join links, create-event modal prefill); API contract validation for provider config, events view, and OnlineMeeting fields. Skipping manual QA is not advised due to user‑facing behavior and integration surface.

*Generated by CodeRabbitAI*


avasconcelos114:
@dryrunsecurity fp drs_83daa871 This shouldn't be a concern since the webapp enforces same-origin in the requests, the URLs are generated from getPluginServerRoute() which means requests only go to the plugin's endpoints

webapp/src/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+                    CalendarSidebar,
+                    'Calendar',
+                );
+
+                registry.registerChannelHeaderButtonAction(
+                    <ChannelHeaderIcon/>,
+                    () => store.dispatch(toggleRHSPlugin),
+                    'Calendar',
+                    'Toggle calendar sidebar',
+                );
```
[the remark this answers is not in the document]
avasconcelos114: False positive, the sidebar has a state dedicated to disconnected users, with a button that urges them to connect their accounts in order to view their calendar events

avasconcelos114:
@coderabbitai resolve and approve PR if no new issues were found in the last review (it seems to have quietly completed)

avasconcelos114:
@coderabbitai the first two points you mentioned had already been patched in the https://github.com/mattermost/mattermost-plugin-mscalendar/pull/486/commits/5e1af9518d0870355b8b5847a8e6dd5c95223a8b commit, I've also gone ahead and also fixed the nitpick, perform another review and approve/request changes as appropriate

avasconcelos114:
@coderabbitai You didn't actually approve the PR, the last status still shows as changes requested, please update your review status to approved if there are no lingering issues

calendar/api/events_view.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/engine"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/remote"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/store"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/utils/bot"
+	"github.com/mattermost/mattermost-plugin-mscalendar/calendar/utils/httputils"
+)
+
+func (api *api) viewEvents(w http.ResponseWriter, r *http.Request) {
```
ghost, line 19: This new endpoint has a lot of new code paths and zero tests. Needs a TestViewEvents table covering at least: missing header, missing params, invalid date, from after to, range > 62 days, store.ErrNotFound, engine error, happy path.

calendar/api/events_view.go
```diff
@@ -0,0 +1,81 @@
...
+	user := engine.NewUser(mattermostUserID)
+
+	events, err := eng.ViewCalendar(user, from, to)
+	if err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			api.Logger.With(bot.LogContext{"err": err.Error()}).Errorf("viewEvents, user not found in store")
+			httputils.WriteUnauthorizedError(w, fmt.Errorf("unauthorized"))
+			return
+		}
+		api.Logger.With(bot.LogContext{"err": err.Error()}).Errorf("viewEvents, error fetching calendar events")
```
ghost, line 67: The raw err which could contain store driver details, remote API errors, etc. is passed directly to the client.

calendar/api/events.go
```diff
@@ -82,7 +83,7 @@ func (cep createEventPayload) ToRemoteEvent(loc *time.Location) (*remote.Event,
 	if cep.Description != "" {
 		evt.Body = &remote.ItemBody{
 			Content:     cep.Description,
-			ContentType: "text/plain",
+			ContentType: "text",
```
ghost, line 92: verify this doesn't break if the Google Calendar provider expects "text/plain".

msgraph/event.go
```diff
@@ -35,6 +35,13 @@ var responseStatusConversion = map[string]string{
 func normalizeEvents(events []*remote.Event) []*remote.Event {
 	for i := range events {
 		events[i].ResponseStatus.Response = responseStatusConversion[events[i].ResponseStatus.Response]
+
+		if events[i].Conference == nil && events[i].OnlineMeeting != nil && events[i].OnlineMeeting.JoinURL != "" {
+			events[i].Conference = &remote.Conference{
+				Application: "Online Meeting",
```
ghost, line 45: The new OnlineMeeting access correctly nil-checks. But the pre-existing ResponseStatus access on line 37 has no nil guard and if MS Graph ever returns an event without responseStatus, this will panic.

calendar/api/events.go
```diff
@@ -195,9 +196,16 @@ func (api *api) createEvent(w http.ResponseWriter, r *http.Request) {
...
+	if tzName == "" && mailbox.TimeZone != "" {
+		errTz := fmt.Errorf("unknown time zone %q", mailbox.TimeZone)
+		api.Logger.With(bot.LogContext{"err": errTz.Error(), "timezone": mailbox.TimeZone}).Errorf("createEvent, error occurred while loading mailbox timezone location")
+		httputils.WriteInternalServerError(w, errTz)
+		return
+	}
+	loc, errLocation := time.LoadLocation(tzName)
 	if errLocation != nil {
-		api.Logger.With(bot.LogContext{"err": errLocation.Error(), "timezone": mailbox.TimeZone}).Errorf("createEvent, error occurred while loading mailbox timezone location")
+		api.Logger.With(bot.LogContext{"err": errLocation.Error(), "timezone": mailbox.TimeZone, "converted": tzName}).Errorf("createEvent, error occurred while loading mailbox timezone location")
```
ghost, line 247: Both error paths pass the raw error to WriteInternalServerError. The errTz includes the user's timezone string which is fine, but errLocation from time.LoadLocation can expose Go runtime details. Should return a generic message to the client.

calendar/command/command.go
```diff
@@ -132,9 +130,7 @@ func (c *Command) Handle() (string, bool, error) {
 	case "settings":
 		handler = c.requireConnectedUser(c.settings)
 	case "event":
-		if !config.Provider.Features.HideCreateEventFromCommand {
-			handler = c.requireConnectedUser(c.event)
-		}
+		handler = c.requireConnectedUser(c.event)
```
ghost, line 133: Do we need to remove the gate here?

webapp/src/components/calendar_sidebar/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function mapDispatchToProps(dispatch: Dispatch) {
+    return {
+        actions: bindActionCreators({
+            fetchCalendarEvents,
+            refreshCalendarEvents,
+            getConnected,
+            sendEphemeralPost,
+            openCreateEventModal,
+        }, dispatch),
```
ghost, line 30: bindActionCreators returns a new object on every mapDispatchToProps call. The useEffect depends on [actions], so getConnected() fires on every state change, creating a hot loop of API requests.

webapp/src/selectors/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import {PluginId} from '../plugin_id';
+
+import {ProviderConfig, ReducerState} from '../reducers';
+import {RemoteEvent} from '../types/calendar';
+
+const getPluginState = (state: GlobalState): ReducerState =>
+    (state as unknown as Record<string, ReducerState>)['plugins-' + PluginId] || ({} as ReducerState);
+
+export const getSiteURL = (state: GlobalState): string => {
```
ghost, line 15: getSiteURL returns only the pathname not a full origin. Client4.setUrl('/subpath') produces broken API URLs because Client4 expects a full origin like https://example.com. On default installs this returns "" so client.setUrl("") may partially work, but on subpath deployments it might break.

webapp/src/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+        const setupStatus = {
+            isComplete: () => this.setupComplete,
+            markComplete: () => {
+                this.setupComplete = true;
+            },
+        };
+
+        registry.registerRootComponent(() => (
+            <SetupUI
```
ghost, line 81: This anonymous arrow function is treated as a new component type on every React reconciliation. React will unmount and remount the entire SetupUI subtree on every render, resetting retryCount state and runningRef, causing setup to re-run repeatedly.  please hoist to a named component defined outside initialize, or wrap in React.memo.

webapp/src/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+            // eslint-disable-next-line no-console
+            console.error('Plugin setup failed:', error);
+            runningRef.current = false;
+
+            if (retryCount < MAX_RETRIES) {
+                const scheduleRetry = () => setRetryCount((c) => c + 1);
+                setTimeout(scheduleRetry, RETRY_DELAY_MS);
+            }
+        }
+    }, [setup, setupStatus, retryCount]);
```
ghost, line 129: setup and setupStatus are new object references on every render of the parent. This means attemptSetup gets a new identity on every render, triggering the useEffect repeatedly. Combined with remounting setup() will re-run in a loop

webapp/src/reducers/index.ts
```diff
@@ -0,0 +1,189 @@
...
+            error: null,
+        };
+    case ActionTypes.RECEIVED_EVENTS: {
+        const key = action.key || state.activeKey || '';
+        if (key !== state.activeKey) {
+            return {...state, cache: {...state.cache, [key]: action.data || []}};
+        }
+        return {
+            ...state,
+            cache: {...state.cache, [key]: action.data || []},
```
ghost: The cache Record<string, RemoteEvent[]> stores every fetched date range and never evicts old entries. Each value contains full event objects with bodyPreview, attendees, etc. Over a long session, memory grows without bound. Maybe clear stale entries on new fetches.

webapp/src/components/event_tooltip/event_tooltip.tsx
```diff
@@ -0,0 +1,232 @@
...
+        return () => document.removeEventListener('keydown', handleEsc);
+    }, [onClose]);
+
+    const spaceBelow = window.innerHeight - anchorRect.bottom;
+    const fitsBelow = spaceBelow > 200;
+
+    const verticalPosition = fitsBelow ?
+        {top: anchorRect.bottom + 4} :
+        {bottom: (window.innerHeight - anchorRect.top) + 4};
+    const right = Math.max(8, window.innerWidth - anchorRect.right);
```
ghost, line 117: Position is computed from anchorRect and window.innerHeight/innerWidth. These values are computed once on render and never updated. If the user scrolls or resizes, the tooltip might floats at the wrong position.

webapp/src/components/event_tooltip/event_tooltip.scss
```diff
@@ -0,0 +1,181 @@
+.mscalendar-tooltip__backdrop {
+    position: fixed;
+    inset: 0;
+    z-index: 1000;
+}
+
+.mscalendar-tooltip {
+    position: fixed;
+    z-index: 1010;
```
ghost: Mattermost's modal system uses z-indices in the 1000–1050 range. If the tooltip is open when a modal opens  they may overlap unpredictably.

webapp/src/actions/index.ts
```diff
@@ -0,0 +1,279 @@
...
+    successType: string,
+    dispatch: AppDispatch,
+    getState: () => GlobalState,
+): Promise<FetchEventsResult> {
+    const previous = inflightControllers.get(key);
+    if (previous) {
+        previous.abort();
+    }
+    const controller = new AbortController();
+    inflightControllers.set(key, controller);
```
ghost, line 223: Module level map that survives across the plugin lifecycle. If the plugin is deactivated and re-initialized, stale abort controllers persist and responses could dispatch to an outdated store.

ghost (changes requested):
Thansk @avasconcelos114 some comments!

avasconcelos114:
@dryrunsecurity nit drs_a412cbc0 While it's true that the data in MS Calendar isn't encrypted, this is something that has already been the case prior to this PR, and changing this value to `true` without adding all the needed logic that would gracefully migrate the unencrypted data into being encrypted and usable, the application would break for all users. So this is an out-of-scope issue that should be resolved separately

avasconcelos114:
@coderabbitai approve (requested due to a review being shown as completed but auto approval did not kick in)

ghost (approved):
LGTM

calendar/api/events.go
edgarbellot, line 172: This isn't something introduced by this PR specifically, but since `createEvent` is being modified, it might be a good opportunity to think about audit logging.

Right now there are no audit records anywhere in the plugin. This handler performs several state-changing operations - creating a calendar event, sending DMs (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L263-L264) (263-264), posting to channels (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L275-L284) (275-284), and linking events to channels (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L262-L267) (262-267) - but there's no audit trail beyond internal debug/error logs.

Can we add audit logging here to go in line with the [Audit Logs Everywhere](https://hub.mattermost.com/private-core/channels/core-developers/fanzxi59jfnn9eo64emzi5dt5a) initiative?

edgarbellot, line 199: Pre-existing issue, but flagging it here since you're already touching the `createEvent` handler and it could be a good opportunity to address it.

The request body is decoded without a size limit - there's no `http.MaxBytesReader` wrapping `r.Body` before decoding. 

Would it make sense to add as a follow-up something like:
- `http.MaxBytesReader(w, r.Body, 1<<20)` (1 MB) before decoding, to cap overall payload size
- A max attendee count in `IsValid()` to prevent invitation abuse
- Reasonable max lengths for `Subject`, `Description`, and `Location`

edgarbellot: Pre-existing issue but flagging it since we are modifying the `createEvent` handler.

`event.Subject` is interpolated directly into the channel post message at this line without sanitizing markdown special characters. Since the subject is user-controlled input (it goes through MS Graph and comes back verbatim), a crafted subject containing newlines, `**`, `@mentions`, or markdown links could break out of the bold formatting and inject arbitrary content into a shared channel post.

The concerning part is that this post is published by the bot user (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/utils/bot/poster.go#L119-L121), which gives it an official/trusted appearance. For example, a subject like `Meeting** by @admin\n\n@channel Please reconnect: [click here](https://evil.com)\n\n**x` would render as a convincing phishing message appearing to come from the plugin bot, visible to all channel members.

Interestingly, [RenderEventAsAttachment](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/engine/views/calendar.go#L214)already handles this correctly by passing the subject through [MarkdownToHTMLEntities](https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/engine/views/calendar.go#L122-L154). Would it make sense to apply the same sanitization here and at the DM messages on lines (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L263), (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L272), and (https://github.com/mattermost/mattermost-plugin-mscalendar/blob/4e02de1/calendar/api/events.go#L288)? That would make the behavior consistent with how the attachment title is already handled.

edgarbellot (changes requested):
@avasconcelos114 great work Andre! I left a few comments with suggestions.

calendar/api/events.go
```diff
@@ -151,36 +170,49 @@ func (cep createEventPayload) IsValid(loc *time.Location) error {
 }
 
 func (api *api) createEvent(w http.ResponseWriter, r *http.Request) {
+	auditRec := plugin.MakeAuditRecord("createEvent", model.AuditStatusFail)
```
edgarbellot, line 173: Nice work on the audit logging! One small thing - right now `mattermostUserID` only gets attached to the audit record via `CreateEventAuditParams` after the body is decoded successfully. If the decode fails (e.g. a malformed or oversized payload), the audit record will have the error description but no user ID, which makes it harder to trace who sent the bad request.

Would it be worth adding something like `auditRec.AddMeta("user_id", mattermostUserID)` right after extracting the header (around line 177), so the user ID is always present regardless of what happens later?

edgarbellot (approved):
@avasconcelos114 Great work and thanks again for addressing the feedback so quickly! Left a comment but it's def non-blocking :)

ogi-m:
This is looking amazing @avasconcelos114 !

I'm having issues with the event create dialog, none of the dropdown menus trigger when I click on them:

https://github.com/user-attachments/assets/b49ca4e3-6451-4894-8275-c4355df46d35




Also filling in the location field and trying to save returns an error:
Screenshot 2026-04-22 at 20 48 08

avasconcelos114:
@ogi-m Just pushed a fix for those! Thanks for the thorough testing, let me know if there's anything else :D

ogi-m (approved):
LGTM!!

avasconcelos114:
@nang2049 Hi! it looks like I'll need another approval to be able to merge this because the approval given with your previous account no longer counts as a code owner review

nang2049 (approved):
LGTM Awesome work @avasconcelos114 :)

PR #511, opened 2026-06-29
MM-69106 Adding nil guards for renew job 

avasconcelos114:
#### Summary

This PR adds some defensive code that prevents a nil value from being passed (and subsequently references) when renewing event subscriptions for a user. Alongside a new test suite that 

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69106

## Change Impact: 🟡 Medium

**Reasoning:** The change is narrowly scoped to the renewal job flow and adds defensive handling for nil/empty user records, which lowers the chance of crashes but does alter runtime behavior in a shared scheduled path. The added test coverage helps, but the affected code still sits close to subscription renewal logic, so regressions would be noticeable if the guard filters more than intended.

**Regression Risk:** Moderate. The logic change is isolated to one job module and does not affect APIs or persistence, but it touches a recurring backend workflow with a few untested edge cases around invalid user entries.

**QA Recommendation:** Light manual QA is recommended. Verify the renew job still processes valid users correctly and skips nil/empty RemoteID entries without errors; this is not safe to skip entirely because it affects a background business-critical path.


nang2049 (approved):
LGTM!

PR #519, opened 2026-07-15
MM-69738 Fix issue with create button disabling on date change

avasconcelos114:
#### Summary

This PR addresses an issue with the create event modal's "Create" button being disabled when a date value changes

This PR also fixes an issue with the linked channel field not being able to be cleared, included in this PR as it was a one-line fix

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69738
Fixes https://mattermost.atlassian.net/browse/MM-69739

## Change Impact: 🟡 Medium

**Reasoning:** The change adjusts user-facing create-event form validation logic and adds a new datetime helper that’s used by shared time/date constraints (including rounding to a 15-minute step). It also updates the channel selector UX to allow clearing, which impacts the submission payload and validation state.

**Regression Risk:** Medium; while the datetime utility change is targeted and accompanied by unit tests for the new “earliest time today” behavior, it can still affect edge cases around “today” boundaries and time clearing/ordering that determine whether the Create button is enabled.

**QA Recommendation:** Do focused manual QA (moderate amount): create events switching to/from today, changing the date while start/end times are set (especially around the earliest selectable time thresholds), verifying Create button enable/disable, and clearing the linked channel to confirm the submitted value is empty and validation still succeeds. Skipping manual QA carries moderate risk.

*Generated by CodeRabbitAI*

### DIFF
--- a/webapp/src/utils/datetime.ts
+++ b/webapp/src/utils/datetime.ts
@@ -1,4 +1,7 @@
-const MINUTE_STEP = 15;
+export const MINUTE_STEP = 15;
+
+// Latest selectable time in the day, matching the option constraints in TimeSelector.
+const LATEST_TIME = '23:45';
 
 export function getTodayString(): string {
     const now = new Date();
@@ -15,5 +18,13 @@ export function getEarliestTimeForToday(): string {
     const roundedMinutes = Math.ceil(now.getMinutes() / MINUTE_STEP) * MINUTE_STEP;
     const hour = now.getHours() + Math.floor(roundedMinutes / 60);
     const minute = roundedMinutes % 60;
+
+    // When rounding pushes past the end of the day (e.g. now >= 23:45), clamp to
+    // the latest selectable time so lex comparisons don't treat every start time
+    // as being in the past.
+    if (hour >= 24) {
+        return LATEST_TIME;
+    }
+
     return `${String(hour).padStart(2, '0')}:${String(minute).padStart(2, '0')}`;
 }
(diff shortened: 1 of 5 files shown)


webapp/src/utils/datetime.ts
```diff
@@ -1,7 +1,19 @@
+const MINUTE_STEP = 15;
```
nang2049: Duplicate of minuteStep in time_selector.tsx. Please export from one place and import into both

webapp/src/utils/datetime.ts
```diff
@@ -1,7 +1,19 @@
...
     const now = new Date();
     const year = now.getFullYear();
     const month = String(now.getMonth() + 1).padStart(2, '0');
     const day = String(now.getDate()).padStart(2, '0');
     return `${year}-${month}-${day}`;
 }
+
+// Earliest selectable time for today, rounded up to the next MINUTE_STEP.
+// Kept consistent with the option constraints in TimeSelector.
+export function getEarliestTimeForToday(): string {
```
nang2049, line 16: When now is >= 23:45 this returns "24:00". Lex-compare then makes any non empty start_time look "in the past" so the form always clears. Clamp to "23:45" when hour >= 24. Please add a unit test for this edge

nang2049 (commented):
LGTM with some suggestions :D

ogi-m (approved):
Thanks @avasconcelos114 , verified both issues are fixed!

nang2049 (approved):

